PR #174, opened 2023-01-08
feat: replace ts-prune with knip

### LINKED ISSUE #192: 🚀 Feature: Replace ts-prune with Knip
### Bug Report Checklist

### Overview

#174: [knip](https://github.com/webpro/knip) looks great and does more than just unused code detection. Let's add it in!

### Additional Info

_No response_

webpro:
Noticed the other day (from https://github.com/release-it/release-it/issues/964) that this repo might benefit from Knip, so here I am introducing Knip to find unused files, dependencies and exports.

Docs: https://github.com/webpro/knip/blob/next/README.md. Knip v1 is almost out, this PR holds v1.0.0-beta.9 but the API shouldn't change anymore, so that should not affect this PR.

Some remarks for now:

- In the dependency department, there are still a few false positives that will sort themselves out with new plugins for `vitest` and `npm-package-json-lint`
- It does look like `yargs` is unused and can be removed?
- Seems like knip and ts-prune agree about there being no unused exports... Obviously it's up to you what to do with this.

Would you be interested at all in having Knip in this repo? Happy to answer any questions or feedback you might have.

Fixes #192


.eslintrc.cjs
```diff
@@ -12,7 +12,7 @@ module.exports = {
 	overrides: [
 		{
 			extends: ["plugin:markdown/recommended"],
-			files: ["**/*.{md}"],
+			files: ["**/*.md"],
```
JoshuaKGoldberg, line 15: Heh, nice catch. If you can pull this out into a separate PR I'd be happy to pull it in quickly.

pnpm-lock.yaml
```diff
@@ -3986,6 +4447,42 @@ packages:
...
+    hasBin: true
+    requiresBuild: true
+    dependencies:
+      '@esbuild-kit/esm-loader': 2.5.4
+      '@npmcli/map-workspaces': 3.0.1
+      '@snyk/github-codeowners': 1.1.0
+      chalk: 5.2.0
+      easy-table: 1.2.0
+      esbuild: 0.16.13
+      esbuild-register: 3.4.2_esbuild@0.16.13
```
JoshuaKGoldberg, line 4468: Err, why are these dependencies of `knip`? I'd really prefer not to have to add a _second_ distribution of `esbuild` in order to take it on as a dev dependency.

package.json
```diff
@@ -55,6 +56,7 @@
 		"format": "prettier \"**/*\" --ignore-unknown",
 		"format:write": "pnpm format --write",
 		"lint": "eslint . --max-warnings 0 --report-unused-disable-directives",
+		"lint:knip": "knip",
```
JoshuaKGoldberg, line 59: You'll also want to add a `.github/workflows/knip.yml` file to run this in CI.

JoshuaKGoldberg (changes requested):
Hey again @webpro! Nice to see a tool getting worked on that can holistically tackle unused code/dependency detection. I'm up for trying it out! Especially since it's already spotted `yargs` as an unnecessary dependency. 😄 

But, a few blockers:
* There shouldn't be any false positives
* Let's get it running in CI with a GitHub Action
* You'll want to remove `ts-prune` in this PR too, since it'll now be redundant
* The dependency on an explicit `esbuild` version

webpro:
> Hey again @webpro! Nice to see a tool getting worked on that can holistically tackle unused code/dependency detection. I'm up for trying it out! Especially since it's already spotted `yargs` as an unnecessary dependency. 😄

Great, happy to hear you are interested!

> But, a few blockers:
> 
> * There shouldn't be any false positives
> * Let's get it running in CI with a GitHub Action
> * You'll want to remove `ts-prune` in this PR too, since it'll now be redundant
> * The dependency on an explicit `esbuild` version

Things are sorted out, except for the `esbuild` dependency, since Knip needs to `require/import` configuration files.

webpro:
Just saw that `markdownlint` is now also unused as Knip also requires a custom dependency resolver to find it in `extends`.

Also enabled production mode:

```
$ knip --production
Unused exports in namespaces (1)
greet  src/greet.ts
```

If interested, read more at https://github.com/webpro/knip/tree/next#production-mode

JoshuaKGoldberg (changes requested):
Ok this is looking great, thanks @webpro! I looked at this more closely and think this is the last list of things to request:

* `ts-prune` is still referenced in the README.md. It should be replaced with knip.
    * Would you like the honors of writing the marketing text for it? 😁 
* `knip.jsonc` references the `setup/` directory. But the setup script removes that directory. The setup script should be augmented to change `knip.jsonc` to no longer reference `setup/`

Do you have time to make these changes? Absolutely no worries if not, I should Soon ™️ 🙂

.eslintrc.cjs
JoshuaKGoldberg: Hmm, this is probably what's causing https://github.com/JoshuaKGoldberg/template-typescript-node-package/actions/runs/3943867603/jobs/6749189690:

```plaintext
/home/runner/work/template-typescript-node-package/template-typescript-node-package/README.md
  96:1  error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/README.md/2_2.ts` using `parserOptions.project`: <tsconfigRootDir>/tsconfig.eslint.json
```

Should probably split this out into a separate issue & PR. I can take a look at this too if you don't have time. ❤️

webpro:
> Ok this is looking great, thanks @webpro! I looked at this more closely and think this is the last list of things to request:
> 
> * `ts-prune` is still referenced in the README.md. It should be replaced with knip.
>   
>   * Would you like the honors of writing the marketing text for it? 😁

Done :)

> * `knip.jsonc` references the `setup/` directory. But the setup script removes that directory. The setup script should be augmented to change `knip.jsonc` to no longer reference `setup/`

The entry files reference the two `script/setup*.js` files, is that not correct?

> Do you have time to make these changes? Absolutely no worries if not, I should Soon ™️ 🙂

Sure, just pushed the changes.

.eslintrc.cjs
webpro: Maybe it's convenient to leave it in this PR? Eventually it's caused by different backing glob parsers (in ESLint and Knip), but necessary for green lights in this PR.

JoshuaKGoldberg:
> The entry files reference the two script/setup*.js files, is that not correct?

...sort of!

In more detail: the repo contains a `script/` folder that includes a `setup.js` file. That file is meant to be run (`pnpm run setup`) when someone forks the template. "Setup" here refers to assorted repo tasks such as renaming to their repo name, populating GitHub labels, etc.

The setup script is also meant to remove the `setup/` directory and any references to setup scripts. Which it should because there's no need to keep that directory around once a repo has been set up! So, anything referring to the `setup/` directory should be removed as a part of `script/setup.js`

Does that all sound reasonable to you? (This repo is still new and I'm still learning how to document+explain it!)

webpro:
Ah, that makes sense. I've now actually ran the setup. You may want to remove the `chalk` dependency then, too, Knip things it's no longer in use after running the setup :)

Ideally, the `script` entries in `knip.jsonc` are removed during setup? So that the `setup*.js` files are still linted before that.

JoshuaKGoldberg:
> remove the `chalk` dependency then, too

Another win for Knip! 😄 #193 

> Ideally, the `script` entries in `knip.jsonc` are removed during setup? So that the `setup*.js` files are still linted before that.

Yes!

webpro:
Alright, I prefer to leave the rest of the project-specific things and scripts to you. Let's get this PR merged?

JoshuaKGoldberg:
Gif of Selena Gomez saying Let's Do It

JoshuaKGoldberg (approved):
🚀

webpro:
Nice!

JoshuaKGoldberg:
@allcontributors add @webpro for feature tooling

JoshuaKGoldberg:
err

@allcontributors add @webpro for a feature please

JoshuaKGoldberg:
oh, right, that's not a thing

PR #206, opened 2023-01-27
chore(knip): upgrade

webpro:
This PR updates Knip and reduces the config it needs:

- Saw that the Markdown glob in the ESLint config was lost, which is why Knip (or rather fast-glob/micromatch) had an issue with it (it's not so much that Knip doesn't understand ESLint overrides/extends, it's about that glob pattern specifically). Not sure why it got lost in the other PR.
- After the latest Knip updates regarding finding binaries in all kinds of scripts, also release-it hooks are scanned for binaries.


.eslintrc.cjs
```diff
@@ -12,7 +12,7 @@ module.exports = {
 	overrides: [
 		{
 			extends: ["plugin:markdown/recommended"],
-			files: ["**/*.{md}"],
+			files: ["**/*.md"],
```
JoshuaKGoldberg, line 15: Right, I'd removed this from the other PR because it was causing issues - then forgot to file a followup issue to look into it. 😅 

I need to figure out how to reconcile embedded TS snippets in Markdown docs...

```plaintext
/home/runner/work/template-typescript-node-package/template-typescript-node-package/README.md
  96:1  error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/README.md/2_2.ts` using `parserOptions.project`: <tsconfigRootDir>/tsconfig.eslint.json
```

webpro: Right, didn't catch that, sorry. Wow TIL `files: ["**/*.md/*.js"]` to lint `js` fenced code blocks, that's pretty cool :)

JoshuaKGoldberg: #208, fun! 🙃

webpro: Oh my, it's literally creating those files. But adding them to `includes` doesn't seem to help. That's a nasty conflict of plugins then, it seems.

But then the markdown plugin didn't actually do something here in this repo, then?

JoshuaKGoldberg: I'm ... guessing not 😂

webpro:
To fix some other issues I changed the implementation of Knip's esljnt plugin, but that has the nice side-effect that we can upgrade Knip + config here without affecting other things.

webpro:
Had git conflicts, eventually I just rebased only the remaining commit on top of `main`.

JoshuaKGoldberg:
Thanks @webpro! 🙌 

@allcontributors please add @webpro for ideas

PR #270, opened 2023-02-26
feat: build.yml: run local file at time of build

### LINKED ISSUE #234: 🛠 Tooling: Add end-to-end CI job that importing the package works
### Bug Report Checklist

### Overview

https://github.com/JoshuaKGoldberg/ts-api-utils/issues/35 was able to occur because nothing in CI enforced that running the built package actually works. 

Let's add a workflow step to CI that, after `pnpm build`, you can `node lib/index.js` with exit code 0.

### Additional Info

It's probably easiest to put this as one more `run:` step after:

https://github.com/JoshuaKGoldberg/template-typescript-node-package/blob/cd43b8bb91353bcbfabd347a98b4342bd3a0dfcf/.github/workflows/build.yml#L15

garuna-m6:
## PR Checklist

- [x] Addresses an existing open issue: fixes #234 

## Overview

Added a test to run a local file at time of build workflow.


.github/workflows/build.yml
```diff
@@ -9,7 +9,15 @@ on:
...
+      - id: test1
+        run: |
+          pnpm build | sed 's/^/ Build Test : '
+      - id: test2
+        run: |
+          if node ./src/index.ts ; then
+              echo $? | sed 's/^/Imports Test : success : '
+          else
+              echo $? | sed 's/^/Imports Test : failed : '
+          fi
```
JoshuaKGoldberg, line 23: These are some very nifty shell scripts 😄 ... but they don't work on my local Macbook (macOS Monterey 12.5). Which is inconvenient enough to be a blocker given how many people use macOS.

```plaintext
$ pnpm build | sed 's/^/ Build Test : '
sed: 1: "s/^/ Build Test : ": unterminated substitute in regular expression
```

I'd suggest removing the fanciness and sticking to one command (e.g. `pnpm build`) for each test. It's _slightly_ less nice testing output but easier to run & simpler to read in this file. Then we can also remove the `id:` attribute, since they'll be more easily identifiable by the command. Thoughts?

.github/workflows/build.yml
```diff
@@ -9,7 +9,15 @@ on:
...
     steps:
-      - uses: actions/checkout@v3
-      - uses: ./.github/actions/prepare
-      - run: pnpm build
+      - id: test1
+        run: |
+          pnpm build | sed 's/^/ Build Test : '
+      - id: test2
+        run: |
+          if node ./src/index.ts ; then
```
JoshuaKGoldberg, line 19: Did you mean `node ./src/index.js`? Running `node` on a `.ts` file without any additional configuration gives `TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts"`.

.github/workflows/build.yml
```diff
@@ -9,7 +9,15 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    uses: [actions/checkout@v3, ./.github/actions/prepare]
     steps:
-      - uses: actions/checkout@v3
-      - uses: ./.github/actions/prepare
```
JoshuaKGoldberg, line 14: OH!? `uses` can be an array? That is very nifty - would you be open to find-and-replacing _all_ the `github/workflows/*.yml` files in this repo in a separate PR? I like this 😄 thank you for showing it!

JoshuaKGoldberg (changes requested):
A start, thanks!

.github/workflows/build.yml
garuna-m6: I also need to test this out

.github/workflows/build.yml
garuna-m6: I wasn't sure about this one, which file should we run at the time of build ? I can alter this accordingly

.github/workflows/build.yml
garuna-m6: ok sounds good will try to fix it this weekend and send for re- review

.github/workflows/build.yml
JoshuaKGoldberg: `node ./src/index.js`. That's the JavaScript file created by running TypeScript (`pnpm build`).

garuna-m6:
@JoshuaKGoldberg  does this look good. Also shall I create a separate issue for the using arrays

.github/workflows/build.yml
```diff
@@ -9,7 +9,9 @@ on:
...
-      - run: pnpm build
+      - id: test
+        run: |
+          pnpm build
+          node ./src/index.js
```
JoshuaKGoldberg, line 17: There's no need to combine these commands into a single `run` with an `id`. Just have two `run`s with one command each. Then the title for each `run` will be the command!

```suggestion
      - id: test
        run: pnpm build
        run: node ./src/index.js
```

.github/workflows/build.yml
JoshuaKGoldberg: https://github.com/JoshuaKGoldberg/template-typescript-node-package/actions/runs/4396405450/workflow gives me the impression it's not actually allowed.

> You have an error in your yaml syntax on line 12

JoshuaKGoldberg (changes requested):
Very close!

.github/workflows/build.yml
JoshuaKGoldberg: @garuna-m6 I think `uses` isn't allowed to be an array. The `build` workflow isn't running. 😞

garuna-m6: Sorry about this, I remember reading somewhere we can use arrays and not able to locate the resource. Making changes against to fix it

.github/workflows/build.yml
```diff
@@ -5,6 +5,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/prepare
       - run: pnpm build
+      - run: node ./src/index.js
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Tip: when working on changes, I'd recommend:

* Reading through this repo's [`.github/DEVELOPMENT.md`](https://github.com/JoshuaKGoldberg/template-typescript-node-package/blob/main/.github/DEVELOPMENT.md) (and in general, when working in any repo, reading through its contributing/development docs).
* Trying things out locally to see whether they work. Although you can't (yet?) run GitHub Actions locally, you can run these `run: ...` scripts to see what they do.

JoshuaKGoldberg:
😄 no worries! I'm just happy it's merged!

PR #271, opened 2023-02-27
feat: hide skip-api prompt by default

### LINKED ISSUE #184: 🚀 Feature: Setup script should hide the --skip-api option (not prompt for it)
### Bug Report Checklist

### Overview

Right now, when you run `pnpm setup` with no arguments, it brings you through a list of options:

```plaintext
✔ What will the kebab-case name of the repository be? · should-skip-semantic-release
✔ What will the Title Case title of the repository be? · Should Skip Semantic Release
✔ What owner or user will the repository be under? · JoshuaKGoldberg
✔ How would you describe the new package? · Small utility to check whether a semantic release should be skipped for a commit.
✔ Whether to skip calling the GitHub API (effectively making this a local-only change). · 
```

That last one, known internally as `--skip-api` is pretty useless to most users - and confusing to anybody who isn't super deeply contextualized in this repo. `--skip-api` is really just for testing during development. I think the option shouldn't even be shown in the CLI. It should just be mentioned in the README.md.

### Additional Info

_No response_

TAKANOME-DEV:
## PR Checklist

- [x] Addresses an existing open issue: fixes #184 

## Overview

Hide the prompt to whether or not calling the github api by setting the value of `skip-api` to true by default.


script/setup.js
```diff
@@ -36,6 +36,8 @@ try {
 		strict: false,
 	});
 
+	values["skip-api"] = true;
```
JoshuaKGoldberg, line 39: Aha! This is a nifty way to solve the issue 😄 nice one.

At first I was -1 on this because it means `--skip-api false` isn't supported... but it looks like that's already the case!

Another issue though, `Pre-filling skip-api to true.` gets logged to the console even if someone provides `--skip-api` in the terminal. Not ideal.

Suggestion: I think it's fine to not use `getPrefillOrPromptedValue`. Directly calling to `prompt` or similar is probably fine. What do you think?

JoshuaKGoldberg (changes requested):
Close! Just a little bug, I think. Thanks for sending this!

script/setup.js
TAKANOME-DEV: Oh yeah, sorry for the late response.
If removing the console log is not a big deal, maybe we can go that route :thinking:

JoshuaKGoldberg: @TAKANOME-DEV just checking in, is this still something you have time for?

JoshuaKGoldberg (commented):
Eeeeeven closer! 😄 thanks for working on it some more!

script/setup.js
JoshuaKGoldberg: @TAKANOME-DEV I just pushed this fix now so I can share this package with some folks - cheers!

JoshuaKGoldberg (approved):

TAKANOME-DEV:
Phew, I was getting an error after removing the `const skipApi` and I didn't know what to do.
You save me before I ask for help. Thanks 👍

PR #308, opened 2023-03-20
chore: update yaml lint config

### LINKED ISSUE #292: 🛠 Tooling: Configure ruleset of yaml linter
### Bug Report Checklist

### Overview

We've recently added https://github.com/ota-meshi/eslint-plugin-yml to the repo and configured rules related to key sorting.
The plugin provides more rules though, we can decide whether some of them are worth enabling:

https://ota-meshi.github.io/eslint-plugin-yml/rules/

- all `yml/no-empty-*` rules look useful
- consistency:
  - `yml/file-extension`
  - `yml/plain-scalar`
  - `yml/block-sequence`
  - `yml/block-sequence-hyphen-indicator-newline`
  - `yml/block-mapping`
  - `yml/block-mapping-question-indicator-newline`

addition of `plugin:yml/recommended` could be a good base to start from

### Additional Info

_No response_

nowyDEV:
## PR Checklist

- [x] Addresses an existing open issue: fixes #292 

## Overview

- Extended yaml linter configuration with new rules to further improve consistency
- Fixed newly reported errors

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -60,14 +60,9 @@ module.exports = {
 		{
 			files: ["**/*.{yml,yaml}"],
 			parser: "yaml-eslint-parser",
-			extends: ["plugin:yml/recommended", "plugin:yml/prettier"],
+			extends: ["plugin:yml/standard", "plugin:yml/prettier"],
 			rules: {
 				"yml/file-extension": ["error", { extension: "yml" }],
-				"yml/block-mapping": "error",
-				"yml/block-mapping-question-indicator-newline": "error",
-				"yml/block-sequence": "error",
-				"yml/block-sequence-hyphen-indicator-newline": "error",
-				"yml/plain-scalar": "error",
 				"yml/sort-keys": [
 					"error",
 					{
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -60,8 +60,14 @@ module.exports = {
 		{
 			files: ["**/*.{yml,yaml}"],
 			parser: "yaml-eslint-parser",
-			extends: ["plugin:yml/base", "plugin:yml/prettier"],
+			extends: ["plugin:yml/recommended", "plugin:yml/prettier"],
 			rules: {
+				"yml/file-extension": ["error", { extension: "yml" }],
+				"yml/plain-scalar": "error",
+				"yml/block-sequence": "error",
+				"yml/block-sequence-hyphen-indicator-newline": "error",
+				"yml/block-mapping": "error",
+				"yml/block-mapping-question-indicator-newline": "error",
 				"yml/sort-keys": [
 					"error",
 					{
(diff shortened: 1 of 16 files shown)


.eslintrc.cjs
```diff
@@ -60,8 +60,14 @@ module.exports = {
 		{
 			files: ["**/*.{yml,yaml}"],
 			parser: "yaml-eslint-parser",
-			extends: ["plugin:yml/base", "plugin:yml/prettier"],
+			extends: ["plugin:yml/recommended", "plugin:yml/prettier"],
```
JoshuaKGoldberg, line 63: Most of the rules enabled below are also enabled in `plugin:yml/standard`. Which I'm only learning about today, reviewing this PR 😅 sorry for suggesting limiting to `recommended` before. Is there any reason not to use `plugin:yml/standard`?

.github/workflows/build.yml
```diff
@@ -10,6 +10,7 @@ name: Build
 
 on:
   pull_request:
+    types: [opened, reopened, synchronize]
```
JoshuaKGoldberg, line 13: 🤔 I like how these are succinct and implicit. Since per https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request these are the defaults, I don't see any need to explicitly declare them... 

I also tried removing the `:` to make line 12 `pull_request` and got a formatting complaint from Prettier:

```plaintext
✖ prettier --ignore-unknown --write:
[error] .github/workflows/build.yml: SyntaxError: Implicit map keys need to be on a single line (12:3)
[error]   10 |
[error]   11 | on:
[error] > 12 |   pull_request
[error]      |   ^^^^^^^^^^^^
[error] > 13 |
[error]      | ^
[error] > 14 |   push:
[error]      | ^^^^^^^
[error]   15 |     branches:
[error]   16 |       - main
[error]   17 |
```

[TIL you can use `~` or `null`](https://stackoverflow.com/questions/34089496/empty-field-in-yaml). It seems to work in #309's builds. Let's go with `~` for all of these files instead of the explicit `types` array?

```suggestion
  pull_request: ~
```

JoshuaKGoldberg (changes requested):
A great start, thanks for sending this @nowyDEV! 🙌

.eslintrc.cjs
nowyDEV: It had couple rules that would conflict with prettier, since we're also using prettier preset we're good to go.

https://github.com/ota-meshi/eslint-plugin-yml/blob/master/src/configs/prettier.ts

.github/workflows/build.yml
nowyDEV: Had to add this explicitly cause the linter was complaining about empty value, the `~` trick does the job  :1st_place_medal:

.eslintrc.cjs
JoshuaKGoldberg: Woop!

JoshuaKGoldberg (approved):
Awesome, thanks @nowyDEV! This is great 😄 I've always wanted this package to have everything linted!

PR #602, opened 2023-07-11
build: update knip to version that detects typescript-eslint@v6 configs

JoshuaKGoldberg:
## Overview

Porting in https://github.com/webpro/knip/commit/7fc31dcae109599e5ff9e51ea056bb03e88cbc81 - thanks for the blazingly fast response time & fix @webpro! ❤️ ⚡


webpro:
@JoshuaKGoldberg Coincidentally, another fix and a much simpler solution was implemented (and released) the other day: https://github.com/webpro/knip/blob/main/src/plugins/eslint/helpers.ts#L87-L88

In case you have a any comments on that I'm all ears, but I guess `@typescript-eslint/eslint-plugin` is an exception of sorts?

JoshuaKGoldberg:
Oh interesting! Filed https://github.com/webpro/knip/issues/195 - we're not an exception, just relatively less usual.

PR #775, opened 2023-09-03
feat: Add help cli

### LINKED ISSUE #703: 🚀 Feature: Add a --help CLI option
### Bug Report Checklist

### Overview

Now that `npx template-typescript-node-package` is a supported -even recommended- use case, folks will want to be able to run `npx template-typescript-node-package --help`. Let's have it print out, for each of the available CLI options, what they do. Also the package version

Vaguely:

```shell
npx template-typescript-node-package --help

✨ template-typescript-node-package v1.2.3 ✨

A quickstart-friendly TypeScript template with comprehensive formatting, linting, releases, testing, and other great tooling built-in.

--base: Whether to scaffold the repository with:
  - `everything` that comes with the template _(recommended)_
  - `minimum` amounts of tooling, essentially opting out of everything
  - `prompt` for which portions to exclude
--create-repository: Whether to create a corresponding repository on github.com (if it doesn't yet exist)
...

👉 See https://github.com/JoshuaKGoldberg/template-typescript-node-package for more documentation.
```

### Additional Info

See https://github.com/JoshuaKGoldberg/template-typescript-node-package/blob/77c10eb9d8dd384ae8e6417d26b92211185b6487/docs/Options.md for descriptions of existing CLI flags. You can directly copy & paste from there. 

I'd also suggest running `parseArgs` to make sure Options.md isn't missing any options (it's entirely possible I forgot about one...).

Arguments are first parsed here: https://github.com/JoshuaKGoldberg/template-typescript-node-package/blob/0307eba66fe59c3a89a189352fd4797aad37736f/src/bin/index.ts#L34-L40

I think you'll want to add a `--help` arg there with a `-h` shorthand. See https://nodejs.org/api/util.html#utilparseargsconfig for examples of `short`.

dominicduffin1:
## PR Checklist

- [x] Addresses an existing open issue: fixes #703 

## Overview

I've added functionality to print to the terminal help documentation based on the existing docs when the `--help` or `-h` terminal flags are used.

This will run in place of the functionality.

I've marked this as a draft PR as I have a few questions so it is not yet complete, maybe you can assist @JoshuaKGoldberg :

- What kind of test coverage would you be looking for for this functionality?
- What is the best approach to fetch the version number?
- I'm not clear what is meant by "running `parseargs`" to check for missing options (from the issue).

Beyond that, is what I've included and how I've formatted the printout what you were expecting?


src/bin/index.ts
```diff
@@ -39,11 +40,22 @@ export async function bin(args: string[]) {
...
+			},
 			mode: { type: "string" },
 		},
 		strict: false,
 	});
 
+	const help = values.help;
+
+	if (help) {
+		logHelpText("");
```
JoshuaKGoldberg, line 55: > What is the best approach to fetch the version number?

I'd probably try with a combination of:

* The APIs and techniques mentioned in https://blog.logrocket.com/alternatives-dirname-node-js-es-modules to get the path to the package's own `package.json` file, relative to whatever file in `lib/`/`src/` is asking for it
* The existing `readFileSafeAsJson` helper to read in JSON files

src/bin/help.ts
JoshuaKGoldberg, line 1: > What kind of test coverage would you be looking for for this functionality?

Ehh, probably not much 😄. There's no logic in this new `logHelpText` function. Maybe just a unit test in `src/bin/index.ts` that it calls to the new function?

If you end up putting in code that reads the `package.json` then probably something to verify that the new function logs the right version.

JoshuaKGoldberg, line 1: > I'm not clear what is meant by "running parseargs" to check for missing options (from the issue).

I... don't remember for sure 😅. I think I'd generally been thinking that it'd be good to have some logic that makes sure everything mentioned in `allArgOptions.ts` is logged? And/or something that makes sure everything in `allArgOptions` is mentioned in `Options.md`? We can ignore that comment 🤷 sorry for the confusion!

src/bin/help.ts
```diff
@@ -0,0 +1,230 @@
+import chalk from "chalk";
+
+import { logLine } from "../shared/cli/lines.js";
+
+export function logHelpText(version: string): void {
```
JoshuaKGoldberg, line 5: > `version: string`

I suspect it might be better to figure out the version only if `logHelpText` is called. Otherwise the extra work of figuring that out will happen even if the user doesn't need it.

_(ignore this if you end up having the CLI always log the version - see later comment)_

src/bin/help.ts
```diff
@@ -0,0 +1,230 @@
...
+	logLine(chalk.cyan(`✨ template-typescript-node-package v${version} ✨`));
+
+	logLine(
+		chalk.cyan(
+			`
+      A quickstart-friendly TypeScript template with comprehensive formatting, 
+      linting, releases, testing, and other great tooling built-in.
+      `,
+		),
+	);
```
JoshuaKGoldberg, line 17: Thinking on the sentences in this output:

```plaintext
┌  Welcome to create-typescript-app ! 🎉
│
│  ⚠️ This template is early stage, opinionated, and not endorsed by the TypeScript team. ⚠️
│  ⚠️ If any tooling it sets displeases you, you can always remove that portion manually. ⚠️
│
│  ✨ template-typescript-node-package v1.29.30 ✨
│  
      A quickstart-friendly TypeScript template with comprehensive formatting, 
      linting, releases, testing, and other great tooling built-in.
```

Seems a little duplicative to mention the title twice. Though I do like the new emojis 😄.

What do you think about:

* Having the CLI always log the version _(why not?)_
* Removing the second mention of the package name

src/bin/help.ts
```diff
@@ -0,0 +1,230 @@
+import chalk from "chalk";
+
+import { logLine } from "../shared/cli/lines.js";
+
+export function logHelpText(version: string): void {
+	logLine();
+
+	logLine(chalk.cyan(`✨ template-typescript-node-package v${version} ✨`));
```
JoshuaKGoldberg, line 8: > `template-typescript-node-package`

It's `create-typescript-app` now 😄 I renamed it in #728

src/bin/help.ts
JoshuaKGoldberg, line 17: Thinking on the formatting of this output:

```plaintext

│  ✨ template-typescript-node-package v1.29.30 ✨
│  
      A quickstart-friendly TypeScript template with comprehensive formatting, 
      linting, releases, testing, and other great tooling built-in.
      
│
│  Core options:
│  
      --base: Whether to scaffold the repository with:
        • everything: that comes with the template (recommended)
        • minimum: amounts of tooling, essentially opting out of everything
        • prompt: for which portions to exclude
```

Screenshot of the CLI output of the previous code block

Note the lack of `|`s on the left. It feels weird to have them only inconsistently. I'm maybe 75% in favor of having them always for consistency & snazzy styling vs. maybe 25% in favor of removing them to keep the help experience simpler. How do you feel?

src/bin/help.ts
```diff
@@ -0,0 +1,230 @@
...
+			`
+      --skip-restore: Skips the prompt offering to restore the 
+        repository if an error occurs during setup
+      `,
+			`
+      --skip-uninstall: Skips uninstalling packages only used for 
+        setup scripts
+      `,
+		),
+	);
```
JoshuaKGoldberg, line 229: Thinking broadly on the output, it looks roughly like a series of repeated tasks:

* Five sections, each of which has a list of options and an optional warning
* Each option has a name, a description, and a type that's either a primitive (`boolean`/`string`) or a union of strings

Both smells strongly to me of a chance to write a standalone function to automate the behavior. That way they're all consistent. Thoughts?

src/bin/help.ts
```diff
@@ -0,0 +1,230 @@
...
+	logLine(chalk.black.bgGreenBright("Optional options:"));
+
+	logLine(
+		chalk.cyan(
+			`
+      --author (${chalk.cyanBright.bold(
+				"string",
+			)}): Username on npm to publish packages under (by 
+        default, an existing npm author, or the currently logged in npm user, 
+        or ${chalk.cyanBright("owner.")}${chalk.magentaBright("toLowerCase()")})
```
JoshuaKGoldberg, line 82: Out of curiosity, why `magentaBright` and for just the method call? I don't think we need to emphasize it so much.

(is it to match an editor theme?)

src/bin/help.ts
```diff
@@ -0,0 +1,230 @@
...
+			`
+      A quickstart-friendly TypeScript template with comprehensive formatting, 
+      linting, releases, testing, and other great tooling built-in.
+      `,
+		),
+	);
+
+	logLine();
+
+	logLine(chalk.black.bgGreenBright("Core options:"));
```
JoshuaKGoldberg, line 21: [Praise] I do really like how the sections are given these headers. Along with the blank lines & spacing, it feels very clean. Nice! 🧹

src/bin/help.ts
JoshuaKGoldberg, line 229: With all these CLI flags explicitly written out in this file, there are now two sources of truth for what options args are available:

* `src/shared/options/args.ts`: The actual runtime values allowed
* `src/bin/help.ts`: What gets printed in `-h`/`--help`

...along with a third value if we include `docs/Options.md`. I have a hard enough time remembering to keep the Markdown docs in sync with the runtime values 😂.  Can those two runtime areas be merged, do you think? Like one big shared object?

JoshuaKGoldberg (commented):
Looks like a solid start! Left some comments around refactors, but I'm very into the general visuals you're going with. 🚀

src/bin/index.ts
dominicduffin1: Thank you, I'll try that!

src/bin/help.ts
dominicduffin1: Thank you!

src/bin/help.ts
dominicduffin1: I think always logging the CLI version sounds good!

src/bin/help.ts
dominicduffin1: Just because I used `cyanBright` to emphasize keywords like `string` and I thought to emphasize the snippet as code it should have two-color highlighting. But I'm not attached to any of the formatting if you have different preferences.

src/bin/help.ts
dominicduffin1: I like this thinking, I'll have a look at it!

src/bin/help.ts
dominicduffin1: I'm also in two minds about this but leaning towards removing them at least from the help section. Although it is a nice little extra on the styling it feels like it could break more easily on certain terminal shapes and also the help section will be easier to maintain without them.

dominicduffin1:
@JoshuaKGoldberg I just wanted to give you a progress update:

- I've refactored the help print logic to use functions and merged the source data into the objects in `src/shared/options/args.ts`.
- I've gone with the option of always logging the version number and deduplicated the package name
- To get the version from the `package.json` I couldn't get the safer looking method of using `URL` to work, I've done something that seems to work in development but I'm not confident about it - will keep on looking at this.
- There are some lint issues with `any` types which I need to resolve.

src/bin/help.ts
JoshuaKGoldberg: Yeah I'd rather not emphasize them quite so much. It's quite a big jump in color.

src/bin/help.ts
JoshuaKGoldberg: Yeah agreed. Let's get them out of here! 🔪

src/bin/help.ts
dominicduffin1: Ok! I've edited it so both parts of the code snippet are the same color.

src/bin/help.ts
dominicduffin1: The changes I pushed yesterday should cover this!

dominicduffin1: Thank you! I've gone for this approach now and I think its a lot cleaner, and it dovetails nicely with the merging of runtime areas you mentioned below as well.

src/bin/index.ts
dominicduffin1: I've implemented this based on the URL method!

src/bin/help.ts
dominicduffin1: I've removed all the `|`s from the help output lines, just leaving them at the top where they also work with the rest of the tool's UI.

dominicduffin1:
Hi @JoshuaKGoldberg! I think I've covered everything that came up in this conversation, so I've marked the PR as ready for review. I am of course willing to work on any additional feedback!

I've been getting an error on one of the e2e tests, `pnpm run test:migrate`, both on this branch and on `main`. I'm not sure if I'm running the test wrong or if there's an issue?

```
Oh no! Running the migrate script modified some files:
 - .github/workflows/build.yml
 - .github/workflows/lint-markdown.yml
 - .github/workflows/lint-package-json.yml
 - .github/workflows/lint-packages.yml
 - .github/workflows/lint-spelling.yml
 - .github/workflows/post-release.yml
 - .github/workflows/prettier.yml
 - .github/workflows/tsc.yml

That likely indicates changes made to the repository without
corresponding updates to templates in src/.
```

src/bin/packageJson.test.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, expect, it } from "vitest";
+
+import { getVersionFromPackageJson } from "./packageJson.js";
+
+describe("Get version number from package.json", () => {
+	it("Returns a string representing a version number when given a valid package.json URL", async () => {
```
JoshuaKGoldberg, line 6: [Naming] Nit: this repo normally goes with a slightly different convention...

* `describe`: just the name of the function (pending https://github.com/veritem/eslint-plugin-vitest/issues/247)
* `it`: lowercase _(otherwise this "X when Y" is phrasing great!)_

```suggestion
describe("getVersionFromPackageJson", () => {
	it("returns a string representing a version number when given a valid package.json URL", async () => {
```

src/bin/packageJson.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export async function getVersionFromPackageJson(): Promise<string | undefined> {
+	const path = new URL("../../package.json", import.meta.url);
+	const data = (await readFileSafeAsJson(path)) as PackageWithVersion;
+
+	if (typeof data === "object" && typeof data.version === "string") {
+		return data.version;
+	}
+
+	return undefined;
```
JoshuaKGoldberg, line 15: [Cleanup] Hmm, in what situation would `data` legitimately not be an object with `"version"`? I think this actually would be a reasonable place to throw an error. That way the `bin` function won't need to check if it's defined.

```ts
const data = ... as Partial<PackageWithVersion> | undefined;
```

What do you think?

src/bin/index.ts
```diff
@@ -8,19 +8,25 @@ import { initialize } from "../initialize/index.js";
...
+
 	prompts.intro(
 		[
-			chalk.greenBright(`Welcome to`),
+			chalk.greenBright(`✨ Welcome to`),
 			chalk.bgGreenBright.black(`create-typescript-app`),
-			chalk.greenBright(`! 🎉`),
+			chalk.greenBright(
+				`${typeof version === "string" ? ` ${version}` : ""}! ✨`,
+			),
```
JoshuaKGoldberg, line 29:
```plaintext
┌  ✨ Welcome to create-typescript-app  1.29.39! ✨
```

[Cleanup] Looks like there's an extra space before the version number?

src/bin/packageJson.test.ts
```diff
@@ -0,0 +1,15 @@
...
+		const version = await getVersionFromPackageJson();
+
+		expect(version).toBeTypeOf("string");
+
+		if (typeof version === "string") {
```
JoshuaKGoldberg, line 11: [Cleanup] https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/no-conditional-expect.md - I'm surprised it's not triggering here. Maybe a bug to file somewhere?

If the `if` is just to make sure `version` is a string, you can use an `as` assertion instead.

That being said, I think you can actually use `import` a little more freely in tests. Vitest massages some runtime stuff nicely for us so we don't have to worry about path relativity or whatnot.

```ts
import packageData from "../../package.json";
```

```ts
expect(version).toBe(packageData.version);
```

src/bin/index.ts
```diff
@@ -39,11 +45,22 @@ export async function bin(args: string[]) {
...
+				type: "boolean",
+			},
 			mode: { type: "string" },
 		},
 		strict: false,
 	});
 
+	const help = values.help;
+
+	if (help) {
```
JoshuaKGoldberg, line 59: [Style] Nitpick: this is only used once, any particular reason to store in a variable?

```suggestion
	if (values.help) {
```

src/bin/help.ts
```diff
@@ -0,0 +1,182 @@
...
+      `,
+		),
+	);
+
+	for (const section of helpTextSections) {
+		logHelpTextSection(section);
+
+		logPipelessLine();
+	}
+}
```
JoshuaKGoldberg, line 182:
```plaintext
┌  ✨ Welcome to create-typescript-app  1.29.39! ✨
│
│  ⚠️ This template is early stage, opinionated, and not endorsed by the TypeScript team. ⚠️
│  ⚠️ If any tooling it sets displeases you, you can always remove that portion manually. ⚠️
 

      A quickstart-friendly TypeScript template with comprehensive formatting, 
      linting, releases, testing, and other great tooling built-in.
      
 
   Core options:

      --base (string): Whether to scaffold the repository with:
        • everything: that comes with the template (recommended)
        • minimum: amounts of tooling, essentially opting out of everything
        • prompt: for which portions to exclude

      --create-repository: Whether to create a corresponding repository on 
        github.com (if it doesn't yet exist)
```

Screenshot of the previous code block in a light mode terminal

Screenshot of the previous code block in a dark mode terminal

[Cleanup] A few thoughts on the display here:

* The `│` / `┌` piping is inconsistent and IMO looks a little off as a result. I think that within a single session, it should either always use it or never.
* The rest of lines are super indented but don't need to be. If there's no `│` piping then let's not add those extra indents?
* What do you think about bolding the options (here, `--base`) to make them stand out?
* I don't see `chalk.cyanBright` any differently (Mac, VS Code, zsh). Maybe `blueBright` instead to make them stand out more?
* After the lists of options, maybe it could log something like `See more documentation at https://github.com/JoshuaKGoldberg/create-typescript-app. 💖`? Some way to encourage folks to visit the repo?

src/bin/help.ts
```diff
@@ -0,0 +1,182 @@
...
+	type: string;
+}
+
+function logHelpTextSection(section: HelpTextSection): void {
+	logPipelessLine();
+
+	logPipelessLine(`   ${chalk.black.bgGreenBright(section.sectionHeading)}`);
+
+	for (const subsection of section.subsections) {
+		if (typeof subsection.warning === "string") {
```
JoshuaKGoldberg, line 33: [Style] Just nipticking here, in cases where `""` isn't ever going to come up I normally default to the more succinct:

```suggestion
		if (subsection.warning) {
```

Was using `typeof` an intentional choice?

src/bin/help.ts
```diff
@@ -0,0 +1,182 @@
...
+      --${flag}${
+				type !== "boolean" ? ` (${chalk.cyanBright(type)})` : ""
+			}: ${description}`,
+				),
+			);
+		}
+	}
+}
+
+function createHelpTextSections(options: object): HelpTextSection[] {
```
JoshuaKGoldberg, line 55: > `options: object`

[Cleanup] Well now this why there are so many `as` assertions 😄 `object` is a super vague type. It just guarantees that the value is an object, not any particular properties on it. 

A more TypeScript-happy approach would be to...

1. Make an interface (`ArgOption`?) that includes `description`, `docsSection`, and `type`, then use it here.
2. Have the `docsSection` type be a union like `"core" | "opt-out" | ...`
3. Have `allArgsOptions` be declared as `satisfies Record<string, ArgOption>`

src/bin/help.ts
```diff
@@ -0,0 +1,182 @@
...
+				subheading: `
+      You can prevent the migration script from making some changes on disk 
+      using any or all of the following CLI flags:`,
+				warning: undefined,
+			},
+		],
+	};
+
+	for (const option of Object.keys(options)) {
+		const data = options[option as keyof object] as unknown as Option;
```
JoshuaKGoldberg, line 115: [Cleanup] ✨ 

```suggestion
	for (const [option, data] of Object.entries(options)) {
```

src/bin/help.ts
```diff
@@ -0,0 +1,182 @@
...
+
+		if (data.docsSection === "core") {
```
JoshuaKGoldberg, line 117: Hmm, this is a lot of `if`s. And they all do almost the same thing. Lots of duplicate code here. How about making an object so the right flags array for each option can be looked up dynamically?

```ts
const sectionFlags = {
	core: core.subsections[0].flags,
	// ...
}
```

```ts
sectionFlags[data.docsSection].push({
	description: data.description,
	// ...
});
```

I was able to get rid of about 60 lines or so with this.

src/bin/help.ts
```diff
@@ -0,0 +1,182 @@
...
+				description: data.description,
+				flag: option,
+				type: data.type,
+			});
+		}
+	}
+
+	helpTextSections.push(core, optional, optOut);
+
+	return helpTextSections;
```
JoshuaKGoldberg, line 160: [Cleanup] Any reason to have a `helpTextSections` array? If it's only used to store the things, this could instead:

```ts
	return [core, optional, optOut];
```

JoshuaKGoldberg (changes requested):
Wonderful 🚀! This is shaping up really nicely. Thanks for iterating on it!

I left some cleanup comments - a few are just style nitpicks, and a few are more structural. Please do let me know if you think I'm off base on any of them (or just want to chat alternatives)! ❤️

JoshuaKGoldberg:
> ```md
> - [ ] Addresses an existing open issue: fixes #000
> ```

Don't forget to reference the issue 😄

dominicduffin1:
@JoshuaKGoldberg Sorry I referenced the issue in the wrong section, fixed that now! Thank you for your comments earlier, I'll look through them in more detail later in the week.

src/shared/cli/lines.ts
```diff
@@ -12,3 +12,11 @@ export function logNewSection(line: string) {
 export function makeLine(line: string | undefined) {
 	return [chalk.gray("│"), line].filter(Boolean).join("  ");
 }
+
+export function logPipelessLine(line?: string) {
```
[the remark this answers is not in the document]
dominicduffin1: I thought it might be repo style to have a log line function but yeah, you're right, it can just be `console.log()`.

src/bin/help.ts
dominicduffin1: No, I don't see any reason why the more succinct version won't work!

dominicduffin1:
@JoshuaKGoldberg Sorry there's been a bit of a delay on me resolving the final feedback - I've been caught up with other things but will be picking this up again in the second half of this week.

src/bin/index.ts
dominicduffin1: You're right! Fixed.

src/bin/packageJson.test.ts
dominicduffin1: I've removed it. Now that I throw an error if its not a string it would be irrelevant anyway.

src/bin/help.ts
dominicduffin1: I made some changes along these lines, thank you!

src/bin/help.ts
dominicduffin1: Good point, I've changed it to use that now!

dominicduffin1:
Hi @JoshuaKGoldberg! I think I've covered everything from your cleanup feedback now. I've also pulled in the latest and migrated the newly added args to the format introduced by this PR in `args.ts` (and added the documentation for them). Let me know what you think.

JoshuaKGoldberg:
Great! Reviewing now.

Tip: [re-requesting a PR review (see the last step, 6)](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) will remove the `status: waiting for author` and notify me.

src/bin/help.ts
```diff
@@ -0,0 +1,172 @@
...
+			type: data.type,
+		});
+	}
+
+	return [core, optional, optOut, offline];
+}
+
+export function logHelpText(): void {
+	const helpTextSections = createHelpTextSections(
+		allArgOptions as Record<string, Option>,
```
JoshuaKGoldberg, line 153: [Cleanup] Nit: seems to be unnecessary? Nothing breaks in `pnpm tsc --watch` when I remove it.

```suggestion
		allArgOptions,
```

src/bin/packageJson.test.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { getVersionFromPackageJson } from "./packageJson.js";
+
+describe("getVersionFromPackageJson", () => {
+	it("returns a string representing a version number when given a valid package.json URL", async () => {
+		const version = await getVersionFromPackageJson();
+
+		expect(version).toBeTypeOf("string");
+
+		expect(version.split(".").length).toBe(3);
```
JoshuaKGoldberg, line 11: [Testing] Test files have a bit more flexibility in what they can import from than published files (since we know they're being run in Vitest, not some user's published files). I think it'd be better to do what the migration test does for importing package data:

https://github.com/JoshuaKGoldberg/create-typescript-app/blob/7649696e7f23285f640a2646312b028922d39458/script/migrate-test-e2e.js#L4

That way the test can `expect(version).toBe(packageData.version)` directly.

src/bin/help.ts
JoshuaKGoldberg: @dominicduffin1 bump on this thread?

src/bin/help.ts
JoshuaKGoldberg: [Testing] Un-resolving this to say: `help.ts` has a lot of untested code. Can I request a bit more strongly:

> Maybe just a unit test in src/bin/index.ts that it calls to the new function?

My headspace now is that it might be fine to just have a `toMatchInlineSnapshot` or similar so we can see that it all looks good. Thoughts?

JoshuaKGoldberg (changes requested):
I've got a thread for you 🙂 progress is looking good!

https://github.com/JoshuaKGoldberg/create-typescript-app/pull/775/files#r1328832785

src/bin/packageJson.test.ts
dominicduffin1: Makes sense, I've implemented this!

dominicduffin1:
Hi @JoshuaKGoldberg! I can't seem to reply to the final two comments on your recent thread so I'm putting my answers here:

> @dominicduffin1 bump on this thread?

I've searched the code and I can't seem to find any remaining instances of `logPipelessLine()` so I'm not sure what you mean by this one.

> [Testing] Un-resolving this to say: help.ts has a lot of untested code. Can I request a bit more strongly:

> Maybe just a unit test in src/bin/index.ts that it calls to the new function?

> My headspace now is that it might be fine to just have a toMatchInlineSnapshot or similar so we can see that it all looks good. Thoughts?

Yes, with the iterations we've gone through I agree there is a lot of logic we should test. As the code stands, I'm torn between testing `logHelpText()`, which could create a maintenance headache as the test would need to keep up with changes in the options, or exporting and testing the `createHelpTextSections()` and `logHelpTextSection()` functions directly. Or maybe `logHelpText()` should be made more testable by taking `allArgOptions` as an argument? Do you have an opinion on these approaches?

JoshuaKGoldberg:
> bump on this thread? ... not sure

I'm not sure either 😄 ignore me. I just marked all conversations as resolved and will re-review.

> torn between testing `logHelpText()`, ... or exporting and testing the createHelpTextSections() and logHelpTextSection() functions directly

I'd be satisfied with a single test that calls what is currently named `logHelpText` and asserts on what happens then. Search `mockConsoleLog` for reference there - it's not unheard of in this repo.

src/bin/packageJson.test.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import packageData from "../../package.json" assert { type: "json" };
+import { getVersionFromPackageJson } from "./packageJson.js";
+
+describe("getVersionFromPackageJson", () => {
+	it("returns the current version number when given a valid package.json URL", async () => {
+		const version = await getVersionFromPackageJson();
+
+		expect(version).toBe(packageData.version);
+	});
```
JoshuaKGoldberg, line 11: There's an `if` in the `getVersionFromPackageJson` function, but only one of the two cases is tested. Could you add a unit test for the error case too please?

src/bin/help.ts
```diff
@@ -0,0 +1,172 @@
...
+      --${flag}${
+				type !== "boolean" ? ` (${chalk.cyanBright(type)})` : ""
+			}: ${description}`,
+				),
+			);
+		}
+	}
+}
+
+function createHelpTextSections(
```
JoshuaKGoldberg, line 61: [Question] This function is only called once, is there a reason you wanted to pass an `options` parameter to it? E.g. I can see an argument for it being useful if it's tested directly as you'd suggested could be an option. If this file only ends up having a single snapshot test, though, I think there'd be a nice little code simplification of not having a param or needing the `Option` interface.

src/bin/help.ts
```diff
@@ -0,0 +1,172 @@
...
+	options: Record<string, Option>,
+): HelpTextSection[] {
+	const core: HelpTextSection = {
+		sectionHeading: "Core options:",
+		subsections: [
+			{
+				flags: [],
+				subheading: undefined,
+				warning: undefined,
```
JoshuaKGoldberg, line 70: [Cleanup] Just nitpicking: I'm not a huge fan of having `: undefined` properties when it'd be doable to make the property optional. Any reason not to make these `?: string` in `HelpTextSection`?

Actually, thinking further: `flags` could be made optional too... Just a few `?? []`s or similar instead in `help.ts`. Even less code!

src/bin/help.ts
```diff
@@ -0,0 +1,172 @@
...
+		if (subsection.warning) {
+			console.log(chalk.yellow(subsection.warning));
+		}
+
+		if (subsection.subheading) {
+			console.log(chalk.green(subsection.subheading));
+		}
+
+		for (const option of subsection.flags) {
+			const { description, flag, type } = option;
```
JoshuaKGoldberg, line 48: [Naming] Hmm, the property is called `flags` but the local variable is `option`. That's a little confusing IMO. And it can't be called flag, because the `flag` property is there. We can bypass thinking of a different name by inlining the destructuring:

```suggestion
		for (const { description, flag, type } of subsection.flags) {
```

It does expose a mild naming oddity though - a `Flag` has a `flag`? Maybe, `SectionFlag` for the interface? `SectionOption`? Something like that? _(I'm not confident or picky here)_

src/bin/index.ts
JoshuaKGoldberg, line 1: Starting a new thread about the whole CLI's output - this is what I get locally right now on the latest commit (8abad7cb207e884ec59172aec2c124d8137841ab):

```plaintext
┌  ✨ Welcome to create-typescript-app 1.35.0! ✨
│
│  ⚠️ This template is early stage, opinionated, and not endorsed by the TypeScript team. ⚠️
│  ⚠️ If any tooling it sets displeases you, you can always remove that portion manually. ⚠️
 

      A quickstart-friendly TypeScript template with comprehensive formatting, 
      linting, releases, testing, and other great tooling built-in.
      
 
   Core options:

      --base (string): Whether to scaffold the repository with:
        • everything: that comes with the template (recommended)
        • minimum: amounts of tooling, essentially opting out of everything
        • prompt: for which portions to exclude
```

It looks inconsistent to me how only the first few lines have the `|` piping. IIRC you'd mentioned somewhere a preference for not having the piping in this help mode. IN that case I'd think it best to:

* Switch `src/bin/index.ts`'s `bin` function to only log the normal intro prompt if it isn't running in help mode
* Have this mode reduce indentation a bit, since it doesn't need to account for the piping

Are you not seeing the `│`s? Am I doing a silly?

JoshuaKGoldberg (changes requested):
Progress! 🚀

src/bin/packageJson.test.ts
dominicduffin1: To be honest, I don't know how to test that case, because the `getVersionFromPackageJson` function pulls in the `package.json` and will only throw the error if there is a problem with doing so. I can't think of any expected case where this will happen. Is there a strategy I'm missing here?

src/bin/help.ts
dominicduffin1: I thought it would be easier to follow what was happening in the code if the `options` were pulled in in the exported function. I'm not set on this though, happy to change it if you think its better?

JoshuaKGoldberg: I do :)

src/bin/help.ts
dominicduffin1: I've made `subheading` and `warning` optional. I can't see any reason why we would have a section without `flags` and we are pushing onto all of the empty arrays in the code currently, so I kind of prefer flags to be mandatory.

src/bin/index.ts
dominicduffin1: You're seeing right! I _was_ thinking that it would be better to keep them just for the bit that is shared with the main program, but thinking about it again I think it would be cleaner to go with your suggestion, so I'll make that change!

dominicduffin1:
> I'd be satisfied with a single test that calls what is currently named logHelpText and asserts on what happens then. Search mockConsoleLog for reference there - it's not unheard of in this repo.

Ok, thank you, I'll have a look at some of the others, read up a bit and work on getting something in there for that!

src/bin/help.ts
dominicduffin1: Done!

src/bin/index.ts
dominicduffin1: I've changed this and I think it looks better now!

src/bin/packageJson.test.ts
dominicduffin1: It seems like mocking the function that reads the file so that it doesn't return the real contents of the file passed ought to be the way to go, but I haven't been able to get it to work.

```typescript
vi.mock("../shared/readFileSafeAsJson", async (importOriginal) => {
	await importOriginal();
	return {
		readFileSafeAsJson: vi.fn(() => {
			return {};
		}),
	};
});
```

I've tried a few ways to make something like this force the `getVersionFromPackageJson()` to reject, but it has stubbornly continued to resolve with the version number.

dominicduffin1:
> I'd be satisfied with a single test that calls what is currently named logHelpText and asserts on what happens then. Search mockConsoleLog for reference there - it's not unheard of in this repo.

I've added a snapshot test for this!

dominicduffin1:
Hi @JoshuaKGoldberg! I've covered the remaining points except for the test case for `getVersionFromPackageJson` where it throws an error, which I can't seem to figure out. I've left in the (failing) test for this so you can see what I was trying to do.

dominicduffin1:
I've also excluded the help snapshot from the cspell linter as it was causing it to fail.

src/bin/packageJson.test.ts
JoshuaKGoldberg: I've never used `importOriginal`. Maybe that's causing issues?

dominicduffin1: I've revised it to look more like the example I found in `src/shared/options/createOptionDefaults/readDefaultsFromDevelopment.test.ts` not using `importOriginal` but it is still stubbornly returning the real value rather than rejecting as it should if the function call to `readFileSafeAsJson` was mocked.

```typescript
const mockReadFileSafeAsJson = vi.fn();

vi.mock("../shared/readFileSafeAsJson", () => ({
	get readFileSafeAsJson() {
		return mockReadFileSafeAsJson;
	},
}));

describe("getVersionFromPackageJson", () => {
	it("returns the current version number from the package.json", async () => {
		const version = await getVersionFromPackageJson();

		expect(version).toBe(packageData.version);
	});

	it("throws an error when there is no version number", async () => {
		mockReadFileSafeAsJson.mockResolvedValue({});

		await expect(() => getVersionFromPackageJson()).rejects.toEqual(
			new Error("Cannot find version number"),
		);
	});
});
```

JoshuaKGoldberg: Tell you what - if it's still giving you pain, push up the failing `vi.mock` + `get ...() {` version (i.e. what you just posted) that seems as close to what the existing projects do, and I can take a look. Half the time it's something irritating to find like an extra `../` not being caught.

dominicduffin1: Thank you! It's pushed - this is the path: `src/bin/packageJson.test.ts`. I've added some comments in the file as well, they may be overkill but I thought they might be helpful.

src/bin/packageJson.test.ts
```diff
@@ -0,0 +1,39 @@
...
+import { getVersionFromPackageJson } from "./packageJson.js";
+
+const mockReadFileSafeAsJson = vi.fn();
+
+/* 
+Mock the readFileSafeAsJson function that is called in getVersionFromPackageJson().
+
+Modeled on readDefaultsFromDevelopment.test.ts line 7
+*/
+vi.mock("../shared/readFileSafeAsJson", () => ({
```
JoshuaKGoldberg, line 13: Aha! Found it - this has to be the same as the file import. That includes an extension.

```suggestion
vi.mock("../shared/readFileSafeAsJson.js", () => ({
```

JoshuaKGoldberg, line 13: Aha! Found it - this has to be the same as the file import. That includes an extension.

```suggestion
vi.mock("../shared/readFileSafeAsJson.js", () => ({
```

Also, the first test needs to call `mockReadFileSafeAsJson.mockResolvedValue`. Which makes sense to me: if the test file is trying to mock out `readFileSafeAsJson`, we should never be able to call its original logic. No need to import `package.json` in this test file at all.

cspell.json
```diff
@@ -6,7 +6,8 @@
 		"CHANGELOG.md",
 		"lib",
 		"node_modules",
-		"pnpm-lock.yaml"
+		"pnpm-lock.yaml",
+		"./src/bin/__snapshots__"
```
JoshuaKGoldberg, line 10: I'd rather just have the snapshot be inline. Less configuration to work with.

src/bin/help.test.ts
```diff
@@ -0,0 +1,27 @@
...
+		logHelpText([
+			chalk.yellow(
+				"⚠️ This template is early stage, opinionated, and not endorsed by the TypeScript team. ⚠️",
+			),
+			chalk.yellow(
+				"⚠️ If any tooling it sets displeases you, you can always remove that portion manually. ⚠️",
+			),
+		]);
+
+		expect(mockConsoleLog.mock.calls).toMatchSnapshot();
```
JoshuaKGoldberg, line 25: Just a heads up, the CI fails because chalk's colors are different on GitHub than locally. https://github.com/JoshuaKGoldberg/expect-no-axe-violations/pull/3 is another place I've had to work around it. I don't like that solution though, it's annoying making even more root files & Vitest config.

You can work around it with code like this instead that replaces chalk's with functions with ones that just return strings:

```ts
function makeProxy<T extends object>(receiver: T): T {
	return new Proxy(receiver, {
		get: () => makeProxy((input: string) => input),
	});
}

vi.mock("chalk", () => ({
	default: makeProxy({}),
}));
```

I have on my todo-list making a dedicated package like this.

src/bin/packageJson.test.ts
dominicduffin1: Thank you! How did I miss that 😞

src/bin/help.test.ts
dominicduffin1: Thank you! I've added this code.

cspell.json
dominicduffin1: I can do this of course, but are you sure? Having added all the logic to automate the creation of the help output from the  `argOptions` it feels like a retrograde step to introduce a fully hard-coded string in the test that will have to be edited every time there is a change. Whereas the automatic snapshot file can be updated by pressing `u` in the test runner.

JoshuaKGoldberg: Inline snapshots are updated by pressing `u` too 🙂 otherwise I'd absolutely want the snapshot file.

dominicduffin1: Ah, great! For some reason I assumed that wouldn't be possible 😧 now I see why inline snapshots are much better! I've swapped it out for an inline one.

JoshuaKGoldberg:
@all-contributors please add @dominicduffin1 for code.

> 🤖 Beep boop! This comment was added automatically by [all-contributors-auto-action](https://github.com/marketplace/actions/all-contributors-auto-action).
> Not all contributions can be detected from Git & GitHub alone. Please comment any missing contribution types this bot missed.
> ...and of course, thank you for contributing! 💙

PR #826, opened 2023-09-13
feat: add validation for readOptions with zod

### LINKED ISSUE #784: 🚀 Feature: Validate parseArgs values with Zod
### Bug Report Checklist

### Overview

Right now, `readOptions` has quite a few `as` assertions:

https://github.com/JoshuaKGoldberg/create-typescript-app/blob/bde3d0b1b8bf47dfc245a94d2efddc0179edc07f/src/shared/options/readOptions.ts#L39-L42

This is because:

* Users might run the CLI with all sorts of wacky things (`npx create-typescript-app --base asdf`)
* I've turned off `strict` parsing in the `parseArgs` (https://nodejs.org/api/util.html#utilparseargsconfig)

Even if `strict` were to be turned back on, there would still need to at least be an assertion on `values.base as InputBase`. So I want to have a more complete input validation solution to assert that the values are what's expected.

Let's use my current favorite, [Zod](https://zod.dev).

### Additional Info

Note that the `uninstallPackages` function will need to be updated to remove the `zod` dependency. See where it calls `await removeDependencies`.

nowyDEV:
## PR Checklist

- [x] Addresses an existing open issue: fixes #784

## Overview

Removed type assertions and replaced them with zod validation.


src/shared/options/readOptions.ts
```diff
@@ -35,42 +37,38 @@ export async function readOptions(args: string[]): Promise<OptionsParseResult> {
...
-			| undefined,
-		excludeLintSpelling: values["exclude-lint-spelling"] as boolean | undefined,
-		excludeLintYml: values["exclude-lint-yml"] as boolean | undefined,
-		excludeReleases: values["exclude-releases"] as boolean | undefined,
-		excludeRenovate: values["exclude-renovate"] as boolean | undefined,
-		excludeTests: values["unit-tests"] as boolean | undefined,
-		funding: values.funding as string | undefined,
-		owner: values.owner as string | undefined,
-		repository: values.repository as string | undefined,
+	const options = optionsFromArgsSchema.parse({
```
nowyDEV, line 40: Not sure how to properly add error handling here, I assume we'd like to use `safeParse` here and conditionally return

```ts
return {
	cancelled: true,
	options,
};
```

The problem with that is our `options` type would contain a lot of union type from `values` - `string | boolean | undefined` :thinking:

src/shared/options/readOptions.ts
```diff
@@ -127,13 +125,15 @@ export async function readOptions(args: string[]): Promise<OptionsParseResult> {
...
 
-	const augmentedOptions = await augmentOptionsWithExcludes({
-		...options,
-		author: options.author ?? (await defaults.owner()),
-		email: options.email ?? (await defaults.email()),
-		funding: options.funding ?? (await defaults.funding()),
-		repository,
-	} as Options);
+	const augmentedOptions = await augmentOptionsWithExcludes(
+		optionsSchema.parse({
```
nowyDEV, line 129: Same problem as above, not sure how to add proper error handling here.

src/shared/options/readOptions.ts
```diff
@@ -148,3 +165,68 @@ export async function readOptions(args: string[]): Promise<OptionsParseResult> {
...
+		skipGitHubApi: z.boolean(),
+		skipInstall: z.boolean(),
+		skipRemoval: z.boolean(),
+		skipRestore: z.boolean().optional(),
+		skipUninstall: z.boolean(),
+		title: z.string().optional(),
+	})
+	// This is necessary since we want to require schema fields, but allow values to be undefined
+	// https://stackoverflow.com/questions/71477015/specify-a-zod-schema-with-a-non-optional-but-possibly-undefined-field
+	.transform((o) => ({
```
nowyDEV, line 207: Wasn't sure how to deal with zod making optional fields out of

```ts
z.union([z.string(), z.undefined()]);
```

I assume we don't want to allow non-existent fields while parsing `mappedOptions`.

nowyDEV:
Looks like adding zod breaks migration script, not sure how to deal with this :crying_cat_face:

src/shared/options/readOptions.test.ts
```diff
@@ -0,0 +1,86 @@
...
+	base: "prompt",
+	github: "mock.git",
+	repository: "mock.repository",
+};
+
+vi.mock("./getPrefillOrPromptedOption.js", () => ({
+	get getPrefillOrPromptedOption() {
+		return vi.fn(() => "mock");
+	},
+}));
```
JoshuaKGoldberg, line 45: [Cleanup] There's no need to `get` or `vi.fn` if the mocks aren't being used as mocks anywhere. I've just used that strategy elsewhere because I don't know of a better/cleaner way to provide function mocks to mocked modules (😩).

```suggestion
vi.mock("./getPrefillOrPromptedOption.js", () => ({
	getPrefillOrPromptedOption: () => "mock",
}));
```

(here and elsewhere)

src/shared/options/readOptions.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+vi.mock("./augmentOptionsWithExcludes.js", () => ({
+	get augmentOptionsWithExcludes() {
+		return vi.fn(() => ({ ...emptyOptions, ...mockOptions }));
+	},
+}));
+
+describe("readOptions", () => {
+	it("validates unsupported argument and cancels the function", async () => {
```
JoshuaKGoldberg, line 69: [Naming] This is super nitpicky and I won't mind if you angrily shut this down 😄 but: I generally try to name tests as _X when Y_. I find it more clear for conveying what the _cause_ (_Y_) is and the _effect_ (_X_). Thoughts?

```suggestion
	it("cancels the function when --base is invalid", async () => {
```

(here and the other one too)

src/shared/options/readOptions.ts
```diff
@@ -35,42 +36,55 @@ export async function readOptions(args: string[]): Promise<OptionsParseResult> {
...
+	if (!optionsParseResult.success) {
+		return {
+			cancelled: true,
+			options: Object.entries(mappedOptions).reduce(
+				(acc, [key, value]) => ({
+					...acc,
+					[key]: typeof value === "string" ? undefined : value,
+				}),
+				{},
+			),
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Yeah it's a good question. Both in dev / type system land and for runtime reporting.

For type system land: right now `OptionsParseCancelled` assumes `options: Partial<Option>`. But Zod is pointing out to us that that's not true - and never was. I think changing `OptionsParseCancelled.options` to `object` would be the right step in the type system, to convey that _"we don't know what `options` actually is, other than that it's some object"_. That way there's no need for this `.reduce`.

For runtime reporting: how about printing with [`zod-error`](https://www.npmjs.com/package/zod-error)?

Putting them together: maybe the `OptionsParseCancelled` type should somehow also include the Zod error if there is one? Some massaging of the types so that `bin/index.ts` knows how to print an invalid inputs error from Zod?

src/shared/options/readOptions.ts
JoshuaKGoldberg: You know, now that we're using Zod I don't mind the fields being optional. It's been inconvenient having to make full objects in tests. I think switching all the `| undefined` properties in `interface Options` to be `?` optional would be the cleanest solution here.

JoshuaKGoldberg (changes requested):
Great stuff! Really excited about this - it's a nice and clean solution that removes a _bunch_ of type assertions. 🔥 

Left some comments, but I'm not super picky on any of them. What do you think?

src/shared/options/readOptions.test.ts
nowyDEV: Yep makes sense

src/bin/index.ts
```diff
@@ -50,7 +51,9 @@ export async function bin(args: string[]) {
 		return 1;
 	}
 
-	const { code, options } = await { create, initialize, migrate }[mode](args);
+	const { code, options, zodError } = await { create, initialize, migrate }[
```
JoshuaKGoldberg, line 54: [Style] As a total non-blocking aside, I went back and forth between making an object like `const runners = { create, ... }` in the root of the file or in the function. It might look cleaner here? Not a blocker whichever way you prefer 😄 just raising that I'm not picky.

src/bin/index.ts
```diff
@@ -61,6 +64,12 @@ export async function bin(args: string[]) {
 
 	if (code) {
 		logLine();
+
+		if (zodError) {
+			const validationError = fromZodError(zodError);
+			prompts.outro(chalk.red(validationError));
```
JoshuaKGoldberg, line 70: > `prompts.outro`

[Bug] The terminal output looks a little buggy with two outros:

```plaintext
┌  Welcome to create-typescript-app ! 🎉
│
│  ⚠️ This template is early stage, opinionated, and not endorsed by the TypeScript team. ⚠️
│  ⚠️ If any tooling it sets displeases you, you can always remove that portion manually. ⚠️
│
●  Tip: to run again with the same input values, use: npx create-typescript-app --mode create --base everything --email wat
│
│
└  Validation error: Invalid email at "email"

└  Operation cancelled. Exiting - maybe another time? 👋
```

How about switching to a `logLine`? I think the ideal would be a blank line around the validatione rror:

```plaintext
┌  Welcome to create-typescript-app ! 🎉
│
│  ⚠️ This template is early stage, opinionated, and not endorsed by the TypeScript team. ⚠️
│  ⚠️ If any tooling it sets displeases you, you can always remove that portion manually. ⚠️
│
●  Tip: to run again with the same input values, use: npx create-typescript-app --mode create --base everything --email wat
│
│  Validation error: Invalid email at "email"
│
└  Operation cancelled. Exiting - maybe another time? 👋
```

JoshuaKGoldberg (changes requested):
Looking great! I think the only blocking request on my end is on the `prompts.outro`. Otherwise everything is either praise or a nitpick. Nice! 🔥

src/bin/index.ts
nowyDEV: I think making a separate variable might make it a little bit more readable.

JoshuaKGoldberg (approved):
Super! This is great - really happy about the reduction of assertions in the codebase. Thanks! ✨

Issue #942, opened 2023-10-06
📝 Documentation: Add FAQ entry about adding dual CJS/ESM output


JoshuaKGoldberg:
### Bug Report Checklist

- [X] I have pulled the latest `main` branch of the repository.
- [X] I have [searched for related issues](https://github.com/JoshuaKGoldberg/create-typescript-app/issues?q=is%3Aissue) and found none that matched my issue.

### Overview

This template only has ESM output because that's the future, yippee, hooray. But some packages do need to have dual CJS/ESM output - e.g. packages such as https://github.com/JoshuaKGoldberg/ts-api-utils that needs to be run by ESLint rules (which must be in CJS right now).

It's not that much work to add dual CJS/ESM emit to a package. I don't want to add it to the template, but I do want to remember how to do it - and have a place to explain my & the package's reasoning. So I'll add an FAQ entry on all that.

https://github.com/JoshuaKGoldberg/ts-api-utils/blob/10e11dc288a50f695a882201462e67dc29ac2c93/tsup.config.ts#L7

https://github.com/JoshuaKGoldberg/ts-api-utils/blob/10e11dc288a50f695a882201462e67dc29ac2c93/package.json#L15

### Additional Info

Shoutout to @atilafassina for prompting this by asking about it 😄 

Note to self: I also thought `are-docs-informative` was dual-emit, since it's used in a lint rule. Looks like it might not be after migrating? Maybe I missed a line?

PR #943, opened 2023-10-06
chore: add pnpm workspace to ignore files

KATT:
## PR Checklist

- [ ] Addresses an existing open issue: fixes #000
- [ ] That issue was marked as [`status: accepting prs`](https://github.com/JoshuaKGoldberg/create-typescript-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [ ] Steps in [CONTRIBUTING.md](https://github.com/JoshuaKGoldberg/create-typescript-app/blob/main/.github/CONTRIBUTING.md) were taken

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,4 @@ coverage*
 lib
 node_modules
 pnpm-lock.yaml
+pnpm-workspace.yaml
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,4 @@
 coverage*/
 lib/
 pnpm-lock.yaml
+pnpm-workspace.yaml
--- a/cspell.json
+++ b/cspell.json
@@ -6,7 +6,8 @@
 		"CHANGELOG.md",
 		"lib",
 		"node_modules",
-		"pnpm-lock.yaml"
+		"pnpm-lock.yaml",
+		"pnpm-workspace.yaml"
 	],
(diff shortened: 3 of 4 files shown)


.eslintignore
```diff
@@ -3,3 +3,4 @@ coverage*
 lib
 node_modules
 pnpm-lock.yaml
+pnpm-workspace.yaml
```
JoshuaKGoldberg, line 6: Hmm, the template right now doesn't support monorepos/workspaces, and doesn't include any non-essential files in the `*ignore` files right now. I'm averse to adding in any neighboring / often-likely-to-used files. The list of potential ones is pretty big and it can be hard to precisely define what should or shouldn't be added. And anything that isn't relevant to a user's project is -potentially confusing- extra cruft for the project.

Maybe this should be an FAQ entry for now (sibling issue if so: #942)? Thoughts?

KATT: It adds annoyance to add these ignores and doesn't hurt the setup


I don't like the FAQ-idea, most people will know what to do but it's just annoying to do

JoshuaKGoldberg:
Sorry, I'm going to close this out as not-accepted/wontfix. Since neither this template nor most of my repos use workspaces, I don't want to add the extra lines to the ignore. Thanks for sending though! ❤️

Issue #734, opened 2023-08-31
Tracking issue: Rolling this out to other repositories I touch


JoshuaKGoldberg:
As of end of August 2023, I want to roll out the current state of this template to other repositories I'm a maintainer -or at least a recently strong committer- on. Filing this tracking issue to make it easier to see which ones have had it applied.

I'm going to iteratively:

1. Start sending initial PRs for all these repos
   - As issues come up, I'll file them in the same [Milestone: _Post-Rename Migration_](https://github.com/JoshuaKGoldberg/create-typescript-app/milestone/2) this issue is in
2. Get to the bottom of the applicable repos
3. If there were any issues filed that change the output repository structure, go back to step 1 and repeat
4. Merge the PRs

Status key:

- ✅: Done
- 🏗️: Templated repository; ran / will run `--mode migrate` to upgrade to latest template features
- 📜: Pre-template repository; ran / will run `--mode migrate` to onboard to template
- 🙅: Not applicable; will not use template
- 🛑: Blocked on an external issue or PR

| Status | Repo                                                                                                                      | Issue or PR                                                                 |
| ------ | ------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------- |
| ✅     | [dmnd/dedent](https://github.com/dmnd/dedent)                                                                             | https://github.com/dmnd/dedent/pull/55                                      |
| ✅     | [JoshuaKGoldberg/all-contributors-auto-action](https://github.com/JoshuaKGoldberg/all-contributors-auto-action)           | https://github.com/JoshuaKGoldberg/all-contributors-auto-action/pull/69     |
| ✅     | [JoshuaKGoldberg/all-contributors-for-repository](https://github.com/JoshuaKGoldberg/all-contributors-for-repository)     | https://github.com/JoshuaKGoldberg/all-contributors-for-repository/pull/102 |
| ✅     | [JoshuaKGoldberg/are-docs-informative](https://github.com/JoshuaKGoldberg/are-docs-informative)                           | https://github.com/JoshuaKGoldberg/are-docs-informative/pull/99             |
| ✅     | [JoshuaKGoldberg/console-fail-test](https://github.com/JoshuaKGoldberg/console-fail-test)                                 | https://github.com/JoshuaKGoldberg/console-fail-test/pull/209               |
[...]

Issue #732, opened 2023-08-31
🐛 Bug: Compliance bot isn't posting comments on PRs from forks


JoshuaKGoldberg:
### Bug Report Checklist

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have pulled the latest `main` branch of the repository.
- [X] I have [searched for related issues](https://github.com/JoshuaKGoldberg/create-typescript-app/issues?q=is%3Aissue) and found none that matched my issue.

### Expected

https://github.com/mtfoley/pr-compliance-action is used to verify that PRs match a set of rules. Those include having a title that confirms to conventional commits. If any failures are detected, it should post a GitHub comment on the PR.

### Actual

It's not :( https://github.com/JoshuaKGoldberg/create-typescript-app/pull/731 -> https://github.com/JoshuaKGoldberg/create-typescript-app/actions/runs/6038776422/job/16385877918?pr=731:

```plaintext
Warning: PR Body did not match required format
Error: This PR's title should conform to specification at https://conventionalcommits.org/
Error: Resource not accessible by integration
```

### Additional Info

I think I'd fixed this before. Need to check on that...

JoshuaKGoldberg:
Maybe related: https://github.com/mtfoley/pr-compliance-action/issues/368

PR #992, opened 2023-10-28
fix: account for base when suggesting rerun cmd

### LINKED ISSUE #905: 🐛 Bug: Rerun suggestions don't factor in --base
### Bug Report Checklist

### Expected

Rerun suggestions should use `--base` to suggest what base starting situation to work with.

For example, if started with:

```shell
npx create-typescript-app --base common --mode create -author JoshuaKGoldberg --description asdf --mode create --owner asdf --repository asdf --title asdf
```

...then the rerun suggestion should be something like:

```plaintext
Tip: to run again with the same input values, use: --base common --mode create --author JoshuaKGoldberg --description asdf --owner asdf --repository asdf --title asdf
```

### Actual

Base template (`--base`) is not factored in, so exclusions are printed manually:

```plaintext
Tip: to run again with the same input values, use: npx create-typescript-app --mode create --base common --access public --author JoshuaKGoldberg --description asdf --email-github git@joshuakgoldberg.com --email-npm npm@joshuakgoldberg.com --exclude-compliance true --exclude-lint-deprecation true --exclude-lint-eslint true --exclude-lint-jsdoc true --exclude-lint-json true --exclude-lint-md true --exclude-lint-package-json true --exclude-lint-packages true --exclude-lint-perfectionist true --exclude-lint-regex true --exclude-lint-spelling true --exclude-lint-strict true --exclude-lint-stylistic true --exclude-lint-yml true --mode create --owner asdf --repository asdf true --title asdf
```

### Additional Info

_No response_

GV14982:
## PR Checklist

- [x] Addresses an existing open issue: fixes #905

## Overview

Add additional filter logic to check if an options entry was excluded by the base, and filter it out of the suggested rerun flags, add new test, update failing test with suggestion from jest

(>^u^)><(^u^)><(^u^<)


GV14982:
I am not entirely sure why this is failing compliance checks?

JoshuaKGoldberg:
From the failure linked in the :x:: https://github.com/JoshuaKGoldberg/create-typescript-app/actions/runs/6690991139/job/18177354788?pr=992:

> ```plaintext
> Error: This PR's title should conform to specification at https://conventionalcommits.org/
> ```


`bug:` is a prefix for an issue; `fix:` would be the right title for a PR. A few linked issues that I hope will eventually make the explanations better:

* #732
* https://github.com/mtfoley/pr-compliance-action/issues/368 -> https://github.com/mtfoley/pr-compliance-action/pull/374
* https://github.com/mtfoley/pr-compliance-action/issues/367

cspell.json
```diff
@@ -34,6 +34,7 @@
 		"tsup",
 		"Unstaged",
 		"vitest",
-		"wontfix"
+		"wontfix",
+		"falsey"
```
JoshuaKGoldberg, line 38: [Typo] MDN lists it as `falsy` 😄 https://developer.mozilla.org/en-US/docs/Glossary/Falsy. I'd say let's stick with that?

```suggestion
		"wontfix"
```

src/create/createRerunSuggestion.test.ts
```diff
@@ -93,4 +94,30 @@ describe("createRerunSuggestion", () => {
...
+			'"npx create-typescript-app --mode create --base common --access public --author TestAuthor --description \\"Test description.\\" --directory . --email-github github@email.com --email-npm npm@email.com --exclude-all-contributors true --keywords \\"abc def ghi jkl mno pqr\\" --mode create --owner TestOwner --repository test-repository --skip-github-api true --skip-install true --skip-removal true --title \\"Test Title\\""',
+		);
+
+		const minimum = createRerunSuggestion({
+			...opts,
+			base: "minimum",
+			...getExclusions(options, "minimum"),
+		});
+
+		expect(minimum).toMatchInlineSnapshot(
```
JoshuaKGoldberg, line 119: [Testing] Hmm, this looks like two unit tests rolled into one. Could you split it up please?

Same general reasoning as https://typescript-eslint.io/contributing/local-development#granular-unit-tests.

src/create/createRerunSuggestion.ts
```diff
@@ -30,10 +34,16 @@ export function createRerunSuggestion(options: Partial<Options>): string {
 	};
 
 	const args = Object.entries(optionsNormalized)
+		// Sort so the base is first, then the rest are compared using the `localeCompare` string method
 		.sort(([a], [b]) =>
 			a === "base" ? -1 : b === "base" ? 1 : a.localeCompare(b),
 		)
-		.filter(([, value]) => !!value)
+		// Filter out all entries that have a key in the excluded object or have a falsey value
```
JoshuaKGoldberg, line 41:
```suggestion
		// Filter out all entries that have a key in the excluded object or have a falsy value
```

src/create/createRerunSuggestion.ts
```diff
@@ -30,10 +34,16 @@ export function createRerunSuggestion(options: Partial<Options>): string {
 	};
 
 	const args = Object.entries(optionsNormalized)
+		// Sort so the base is first, then the rest are compared using the `localeCompare` string method
```
JoshuaKGoldberg, line 37: [Docs] Yeah I see why you'd add this comment in 😄 this kind of sorting is not super readable ... mentioning the `localeCompare` does seem a little redundant to me. How do you feel about a reword something like this?

```suggestion
		// Sort so the base is first, then the rest are compared alphabetically
```

JoshuaKGoldberg (changes requested):
Wonderful, thanks! This will make my life a bit easier for #734 😄 

Not much to change, the core logic is looking great. Just a few small touchups. What do you think?

src/create/createRerunSuggestion.test.ts
```diff
@@ -41,6 +42,11 @@ const options = {
 	title: "Test Title",
 } satisfies Options;
 
+const opts = {
```
JoshuaKGoldberg, line 45: [Style] A couple of problems with this `opts` name:

* It's the same as `options`, just shorter. That doesn't convey any new or different meaning. A little confusing having both.
* I generally don't like abbreviated names. It can be hard for folks -especially newcomers and non-native English speakers- to mentally map between the abbreviated and full names.

If there's an explicit need for this second object, then we should name it after that need. But it looks like there isn't - it's just used in the last two added tests. So I think it can be inlined into those two tests. In other words: let's remove the variable and put `excludeLintKnip: undefined` directly in the tests.

What do you think?

JoshuaKGoldberg (approved):
Great, thanks! Just one small nitpick on the `opts` variable name - I can merge in a few days if you don't have strong objection to renaming.

src/create/createRerunSuggestion.test.ts
GV14982: Okay yeah, you make good points there. Thanks for the suggestion! I'll get that pushed up in the morning.

JoshuaKGoldberg:
👋 @GV14982 still interested in this? No worries if not, I just don't want to take over 😄

GV14982:
Oh sorry, I misunderstood your earlier comment. Yeah by all means!

PR #1000, opened 2023-10-30
feat: drop dependency on gh

### LINKED ISSUE #667: 🚀 Feature: Drop end-user dependency on gh
### Bug Report Checklist

### Overview

The [GitHub CLI (`gh`)](https://cli.github.com) is really convenient. I use it a lot personally. Right now, `gh` is used for two things in the repo's hydration/setup scrips:

1. Getting a [GitHub auth token](https://docs.github.com/en/authentication) for the user: search `gh auth` in the code
2. Running commands to modify the repository: search `gh ` _(note the trailing space)_

However, not everyone using this template might be a `gh` user already. It's inconvenient that you'd have to use it. Let's remove the dependency on having that tool installed. I think that'll mean two changes:

1. For the auth token: try running the same `gh auth` scripts as before, and if they fail, fall back to a CLI flag like `--github-auth`
2. Switch to using [Octokit](https://github.com/octokit/octokit.js) instead of `gh ...` for any API calls

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #667

## Overview

Replaces the end-user dependency on `gh` (the GitHub CLI) with more manual Octokit calls. It's a bit less convenient this way but is more type-safe and means users don't need to have some random GitHub utility installed to use the template.

Update December 2023: this works but I don't like how the user has to either have `gh` logged in or use a `process.env.GH_TOKEN`... I'd like to find the time to look into alternatives to log the user in.

Update August 2024: I don't want to procrastinate any more. This PR doesn't make things _worse_. So as long as error messages explicitly tell people to either log in with `gh` or set `process.env.GH_TOKEN`, this is fine.


JoshuaKGoldberg:
For retrieving a user auth token, https://github.com/JoshuaKGoldberg/prune-github-notifications/issues/167 also tracks creating a standalone package.

JoshuaKGoldberg:
The not having a better way than `gh` or a manual `process.env.GH_TOKEN` is irksome, but IMO shouldn't block this from going in. I'll just add in some friendly messaging in the process of retrieving the user's GH auth token.

src/shared/options/createRepositoryWithApi.ts
```diff
@@ -22,14 +22,20 @@ export async function createRepositoryWithApi(
...
+			});
+		} else {
+			await octokit.rest.repos.createInOrg({
+				name: options.repository,
+				org: options.owner,
+			});
+		}
+	} catch (error) {
+		throw new Error("Failed to create new repository on GitHub.", {
+			cause: error,
```
JoshuaKGoldberg, line 38: Not strictly necessary for this PR, but the bad error was bugging me when I was testing out permissions. 😄

Issue #877, opened 2023-09-22
🐛 Bug: Renovate PRs are getting merged with failing migration test


JoshuaKGoldberg:
### Bug Report Checklist

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have pulled the latest `main` branch of the repository.
- [X] I have [searched for related issues](https://github.com/JoshuaKGoldberg/create-typescript-app/issues?q=is%3Aissue) and found none that matched my issue.

### Expected

PRs from bots should never be auto-merged if there's a failing build.

### Actual

https://github.com/JoshuaKGoldberg/create-typescript-app/pull/870 was and broke `main`'s end-to-end migration test job!

https://github.com/JoshuaKGoldberg/create-typescript-app/actions/runs/6269024180/job/17024876442?pr=870

How?!

JoshuaKGoldberg:
Omg. @johnnyreilly you're right - I never set up the end-to-end tests as required for branch protections. Thank you!!

JoshuaKGoldberg:
https://github.com/JoshuaKGoldberg/create-typescript-app/pull/1315#issuecomment-1998354983

> @johnnyreilly I think your suggest worked! This is blocked on the migration script failing. Yay! 🥳 thank you again!

PR #1048, opened 2023-11-21
chore: verify changes on allowedFiles according to the template

### LINKED ISSUE #546: 🧪 Testing: End-to-end hydrate tests should still verify expected changes to files are correct
### Bug Report Checklist

### Overview

Right now, the end-to-end hydrate test allows quite a few files to be changed:

https://github.com/JoshuaKGoldberg/template-typescript-node-package/blob/cf524e64e5fc12ac57f0c46da6050d1b1b7d6137/script/hydrate-test-e2e.js#L26

Many or all of those files should be - e.g. `all-contributorsrc` gets updated with new contributors. But the script just totally ignores those files instead of asserting their changes are what's expected. Which means it's easy to make changes to the repo's versions of files that aren't reflected in the hydration templates for them.

For example, changes to `.eslintrc.cjs` (#528) still pass even though `rootFiles.ts` > `.eslintrc.cjs` isn't updated.

Can we figure out a way to still verify that the changes to those files are what we'd expect them to be?

### Additional Info

Hydration is described here: https://github.com/JoshuaKGoldberg/template-typescript-node-package/blob/ad3d34a3a898f4bb5d8ec7de82409def50c99463/README.md#L92

Aslemammad:
script/migrate-test-e2e.js

## PR Checklist

- [x] Addresses an existing open issue: fixes #546 

## Overview

We track the changes using createRootFiles and we watch the changes files based on the function result.


script/migrate-test-e2e.js
```diff
@@ -8,10 +9,43 @@ const emailGithub = "github@joshuakgoldberg.com";
...
+		.filter(([k]) => !!k),
+);
+
+const rootFiles = await createRootFiles({
+	email: {
+		github: args["--email-github"][1],
+		github: args["--email-npm"][1],
+	},
+	...argsObj,
+});
```
JoshuaKGoldberg, line 43: [Refactor] This is a nifty approach, basing the expected output files on the function that's meant to create them (`createRootFiles`). Unfortunately that sets up something of a circular dependency: that we're setting up the test for expected files on source, but that same source is essentially what we're trying to test! Which is why the bug in the test of `email` having two `github` keys (instead of `github` and `npm`) wasn't caught.

Unfortunately, I don't see a way around having this test file explicitly list out the expected changes. Hopefully they should be pretty minimal and not too many lines per file.

JoshuaKGoldberg (changes requested):
A good start, thanks for looking into this! I think the general strategy is mostly implemented well... but the strategy itself I think is going to need to be reworked. What do you think?

JoshuaKGoldberg:
Btw, looks like there are some CI failures. You'll want to fix them up. `.github/DEVELOPMENT.md` should be helpful in pointing out the tools. If anything's unclear let me knwo!

JoshuaKGoldberg:
Whoops, I pushed to the wrong branch working on another PR - sorry about that :)

script/migrate-test-e2e.js
```diff
@@ -1,5 +1,11 @@
 import chalk from "chalk";
 import { $, execaCommand } from "execa";
+import { createVitest } from "vitest/node";
+
+const vitest = await createVitest("test", {
+	include: [new URL("./verify-changes.test.ts", import.meta.url).pathname],
+	watch: false,
+});
```
JoshuaKGoldberg, line 8: [Praise] Ha, I forgot Vitest has a standalone `createVitest` method. Nifty usage of it!

script/verify-changes.test.ts
```diff
@@ -0,0 +1,21 @@
+import fs from "fs/promises";
+import { expect } from "vitest";
+import { test } from "vitest";
+
+test("knip.jsonc", async () => {
```
JoshuaKGoldberg, line 5: [Refactor] This is a great start! +1 that this test catches the specific properties that are different now.

But: it only catches these specific properties. If the template were to be updated in the future with some other properties, this test wouldn't catch them. For example, let's say a PR adds a `"someAwesomeProperty": true` to the 🍎  _template_ but not the 🍌 `knip.jsonc` file. This test wouldn't fail. But it should fail: it's meant to catch any mismatch between the 🍎 template and the 🍌 file.

I think a more future-proof strategy would be to assert on the actual difference between the 🍎 template and the 🍌 file. How about checking for what the `git diff` output is back in `migrate-test-e2e.js`? The file already has sample code running `git diff HEAD -- ...`.

JoshuaKGoldberg (changes requested):
Progress! 🚀

script/verify-changes.test.ts
Aslemammad: I see, that makes sense! I'm afraid with `git diff HEAD -- ...` it's hard to assert the changes that for instance add a new item to the config array. 

Can we do it through Object.entries or Object.keys also for instance? 

Sorry for the late reply.

script/__snapshots__/verify-changes.test.ts.snap
```diff
@@ -0,0 +1,849 @@
...
++	<a href=\\"https://github.com/JoshuaKGoldberg/create-typescript-app/blob/main/LICENSE.md\\" target=\\"_blank\\"><img alt=\\"License: MIT\\" src=\\"https://img.shields.io/github/license/JoshuaKGoldberg/create-typescript-app?color=21bb42\\"></a>
++	<a href=\\"https://github.com/sponsors/JoshuaKGoldberg\\" target=\\"_blank\\"><img alt=\\"Sponsor: On GitHub\\" src=\\"https://img.shields.io/badge/sponsor-on_github-21bb42.svg\\" /></a>
++	<img alt=\\"Style: Prettier\\" src=\\"https://img.shields.io/badge/style-prettier-21bb42.svg\\" />
++	<img alt=\\"TypeScript: Strict\\" src=\\"https://img.shields.io/badge/typescript-strict-21bb42.svg\\" />
+ 	<img alt=\\"npm package version\\" src=\\"https://img.shields.io/npm/v/create-typescript-app?color=21bb42\\" />
++	<a href=\\"#contributors\\" target=\\"_blank\\"><img alt=\\"All Contributors: 36 🤝\\" src=\\"https://img.shields.io/badge/all_contributors-36_🤝-21bb42.svg\\" /></a>
++	<img alt=\\"Sponsor: On GitHub 💸\\" src=\\"https://img.shields.io/badge/sponsor-on_github_💸-21bb42.svg\\" />
```
JoshuaKGoldberg, line 665: [Non-actionable] Aha, a diff I wasn't expecting!

Filed #1088 to track fixing this unexpected difference. No action needed as part of this PR.

script/__snapshots__/verify-changes.test.ts.snap
```diff
@@ -0,0 +1,849 @@
...
+ 		\\"lib/\\",
+ 		\\"package.json\\",
+ 		\\"LICENSE.md\\",
+@@ -24,8 +23,7 @@
+ 	\\"scripts\\": {
+ 		\\"build\\": \\"tsup\\",
+ 		\\"format\\": \\"prettier \\\\\\"**/*\\\\\\" --ignore-unknown\\",
+-		\\"initialize\\": \\"tsx ./src/bin/index.js --mode initialize\\",
+-		\\"lint\\": \\"eslint . .*js --max-warnings 0\\",
++		\\"lint\\": \\"eslint . .*js --max-warnings 0 --report-unused-disable-directives\\",
```
JoshuaKGoldberg, line 824: [Non-actionable] Aha, a diff I wasn't expecting!

Filed #1089 to track fixing this unexpected difference. No action needed as part of this PR.

script/__snapshots__/verify-changes.test.ts.snap
```diff
@@ -0,0 +1,849 @@
...
+-		\\"test:migrate\\": \\"node script/migrate-test-e2e.js\\",
+ 		\\"tsc\\": \\"tsc\\"
+ 	},
+ 	\\"lint-staged\\": {
+@@ -107,7 +102,7 @@
+ 		\\"vitest\\": \\"^0.34.3\\",
+ 		\\"yaml-eslint-parser\\": \\"^1.2.2\\"
+ 	},
+-	\\"packageManager\\": \\"pnpm@8.10.5\\",
++	\\"packageManager\\": \\"pnpm@8.7.0\\",
```
JoshuaKGoldberg, line 843: [Non-actionable] Lol. Yet another place impacted by #877.

script/__snapshots__/verify-changes.test.ts.snap
```diff
@@ -0,0 +1,849 @@
...
+   \\"commitConvention\\": \\"angular\\",
+   \\"contributors\\": [
+@@ -24,335 +24,253 @@
+       \\"name\\": \\"Jeff Wen\\",
+       \\"avatar_url\\": \\"https://avatars.githubusercontent.com/u/3297859?v=4\\",
+       \\"profile\\": \\"https://sinchang.me\\",
+-      \\"contributions\\": [
+-        \\"code\\"
+-      ]
++      \\"contributions\\": [\\"code\\"]
```
JoshuaKGoldberg, line 22: [Non-actionable] Hmm, after https://github.com/all-contributors/cli/pull/356 and https://github.com/all-contributors/cli/pull/360 I would have thought all-contributors would be respecting the repo's Prettier config... do you have any idea why there are diffs here? I don't think we need to change anything for this PR around formatting. Just curious.

JoshuaKGoldberg (changes requested):
🙌 this is looking great, nice work! I like the snapshot approach. And it's already found a bunch of discrepancies - filed inline to fix separately. Yay!

Requesting changes largely because the script never finishes running, and filtering out a few specific things from the snapshots. I left a few questions inline too.

script/verify-changes.test.ts
JoshuaKGoldberg: Looks like you figured it out in a later commit!

script/__snapshots__/verify-changes.test.ts.snap
```diff
@@ -0,0 +1,849 @@
...
+ 
+ <!-- ALL-CONTRIBUTORS-LIST:END -->
+ <!-- spellchecker: enable -->
++
++<!-- You can remove this notice if you don't want it 🙂 no worries! -->
++
++> 💙 This package was templated with [create-typescript-app](https://github.com/JoshuaKGoldberg/create-typescript-app)."
+`;
+
+exports[`verify bin/index.js 1`] = `""`;
```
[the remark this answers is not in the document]
Aslemammad: Let's try removing it IMO in that case.

script/__snapshots__/verify-changes.test.ts.snap
```diff
@@ -0,0 +1,849 @@
...
+ 		\\"yaml-eslint-parser\\": \\"^1.2.2\\"
+ 	},
+-	\\"packageManager\\": \\"pnpm@8.10.5\\",
++	\\"packageManager\\": \\"pnpm@8.7.0\\",
+ 	\\"engines\\": {
+ 		\\"node\\": \\">=18\\"
+ 	},"
+`;
+
+exports[`verify pnpm-lock.yaml 1`] = `""`;
```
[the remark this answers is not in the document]
Aslemammad: same with this

script/verify-changes.test.ts
```diff
@@ -0,0 +1,10 @@
+import { execaCommand } from "execa";
+import { expect } from "vitest";
+import { test } from "vitest";
+
+import { filesExpectedToBeChanged } from "./constants.js";
+
+test.each([...filesExpectedToBeChanged])("verify %s", async (file) => {
+	const { stdout } = await execaCommand(`git diff HEAD -- ${file}`);
+	expect(stdout).toMatchSnapshot();
```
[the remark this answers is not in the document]
Aslemammad: We can try ignoring these files, or just diff part of those files. The first one would stop verifying which I don't think is the ideal solution, and the second solution might be complicated (AFAIK), which one do you prefer?

script/migrate-test-e2e.js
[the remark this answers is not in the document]
Aslemammad: The reason for it hanging I guess is that we don't exit which is something expected! Let me try exiting.

script/verify-changes.test.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: I'd prefer the latter for completeness yeah. The code could apply some kind of filter to them or instead of using snapshots use like a [`.toMatch`](https://vitest.dev/api/expect#tomatch). But if that's too much for this PR I think it'd be fine to ignore them for now and file a followup issue.

script/__snapshots__/verify-changes.test.ts.snap
```diff
@@ -0,0 +1,849 @@
...
+"diff --git a/package.json b/package.json
+index 7a0e08a..0c382e9 100644
+--- a/package.json
++++ b/package.json
+@@ -8,14 +8,13 @@
+ 	},
+ 	\\"license\\": \\"MIT\\",
+ 	\\"author\\": {
+-		\\"name\\": \\"Josh Goldberg\\",
++		\\"name\\": \\"Aslemammad\\",
```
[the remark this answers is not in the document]
Aslemammad: I tried the --author flag, it seems that it removes the whoe author field. Here's the diff

```diff
-@@ -8,14 +8,13 @@
-       },
-       \\"license\\": \\"MIT\\",
-       \\"author\\": {
--              \\"name\\": \\"Josh Goldberg\\",
-+              \\"name\\": \\"Aslemammad\\",
-               \\"email\\": \\"npm@joshuakgoldberg.com\\"
+@@ -13,9 +13,7 @@
```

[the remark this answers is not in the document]
Aslemammad: let me revert

script/__snapshots__/verify-changes.test.ts.snap
Aslemammad: Haven't found the reason for this yet.

Aslemammad:
Now updating snapshot is available using -u! Plus some files are removed from expected files to change. 

I also tried closing the vitest process after finishing so the CI would not hang. Let me know.

script/__snapshots__/verify-changes.test.ts.snap
JoshuaKGoldberg: Me neither 😄 resolving the thread. A mystery for another day.

script/migrate-test-e2e.js
```diff
@@ -1,7 +1,14 @@
 import chalk from "chalk";
 import { $, execaCommand } from "execa";
+import { createVitest } from "vitest/node";
+
+const vitest = await createVitest("test", {
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Ping @Aslemammad - the `await vitest.start(); await vitest.exit();` seems to work. Nice!

But going a step further, is there anything stopping this file from being a Vitest test itself? Like, being able to run `vitest script/migrate-test/e2e.js` and use `describe`/`expect`/etc. inside this file itself? That way we wouldn't have to manually use the Vitest Node API.

JoshuaKGoldberg (changes requested):
Progress! Looks like there are some failures still though. Requesting changes on getting them resolved. If you have a hard time with any of them, you can always start a thread with questions and I can try to help 🙂.

Everything that happens in CI you can also trigger locally, manually. See `.github/workflows/*.yml` for the exact list of commands that get run. For example, the _**Lint spelling / lint_spelling (pull_request)**_ task comes from `.github/workflows/lint-spelling.yml.`

https://github.com/JoshuaKGoldberg/create-typescript-app/blob/c348a61315beba56f9059ae27541158754e9789d/.github/workflows/lint-spelling.yml#L7

(that `./.github/actions/prepare` command is a local path too, and pretty much all it does is `pnpm install`)

script/migrate-test-e2e.js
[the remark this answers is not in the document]
Aslemammad: My only concern is that running vitest on the whole project would also include this file part of the tests! unless we exclude it and use another vitest command for it! Since this file has so many mutations, this makes me even more concerned. 
As far as I know, the tests in the scripts/ are different from the general tests.

Let me know your opinion if I should keep it or try the new command way.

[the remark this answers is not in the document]
JoshuaKGoldberg: > exclude it and use another vitest command for it!

Yeah you can customize the `-t`/`--testNamePattern` Vitest runs with: https://vitest.dev/guide/cli.html#options. The `test:migrate` command could specify one that only runs this file, not the others.

Aslemammad:
I tried running the CI in my repo, but Github does not trigger it! I wonder if "Cleaning up" should take that much time and therefore the CI stopping the process is all expected.

Aslemammad:
The errors we're currently getting are related to snapshots which I guess (I cannot see the error completely, GitHub CI does not show it) it's because I'm pushing the snapshots that were generated by my system and not the CI. Any idea around that?

script/__snapshots__/migrate-test-e2e.test.js.snap
```diff
@@ -0,0 +1,876 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`verify .all-contributorsrc 1`] = `
+"diff --git a/.all-contributorsrc b/.all-contributorsrc
+index 084b7ee..0ed629e 100644
```
JoshuaKGoldberg, line 5: > `084b7ee..0ed629e`

Having the git commit hash contained in these snapshots means they'll be out of date whenever there's a new commit. I think they'll need to be stripped out.

script/__snapshots__/migrate-test-e2e.test.js.snap
```diff
@@ -0,0 +1,876 @@
...
+--- a/.all-contributorsrc
++++ b/.all-contributorsrc
+@@ -1,5 +1,5 @@
+ {
+-  \\"badgeTemplate\\": \\"<a href=\\\\\\"#contributors\\\\\\" target=\\\\\\"_blank\\\\\\"><img alt=\\\\\\"All Contributors: <%= contributors.length %> 🤝\\\\\\" src=\\\\\\"https://img.shields.io/badge/all_contributors-<%= contributors.length %>_🤝-21bb42.svg\\\\\\" /></a>\\",
```
JoshuaKGoldberg, line 10: > `\\"`

Ah, this looks like it's on an old version of Vitest. This repo recently bumped to their latest major, which included a streamlining of snapshots (https://github.com/vitest-dev/vitest/pull/3961).

Running `pnpm run test:migrate -u` locally should update these.

script/migrate-test-e2e.test.js
```diff
@@ -0,0 +1,73 @@
+import chalk from "chalk";
+import { $, execaCommand } from "execa";
+import { test } from "vitest";
+import { assert, beforeAll, expect } from "vitest";
+
+import packageData from "../package.json" assert { type: "json" };
+import { filesExpectedToBeChanged } from "./constants.js";
+
+beforeAll(async () => {
```
JoshuaKGoldberg, line 9: [Question] Do these _need_ to be in a `beforeAll`? When I run this locally, the logging from `$` gets all mixed up with logging from Vitest. If I remove the `beforeAll` and run the code directly in the file, it all seems to work well.

script/migrate-test-e2e.test.js
```diff
@@ -0,0 +1,73 @@
...
+	const emailNpm = "npm@joshuakgoldberg.com";
+	const owner = "JoshuaKGoldberg";
+	const title = "Create TypeScript App";
+
+	await $({
+		stdio: "inherit",
+	})`c8 -o ./coverage-migrate -r html -r lcov --src src node ./bin/index.js --base everything --mode migrate --description ${description} --email-github ${emailGithub} --email-npm ${emailNpm} --owner ${owner} --title ${title} --repository ${repository} --skip-all-contributors-api --skip-github-api --skip-install`;
+}, Infinity);
+
+test.each([...filesExpectedToBeChanged])("verify %s", async (file) => {
```
JoshuaKGoldberg, line 21: [Praise] I really like the assigning each of these their own test! The more granular reporting is a nice touch.

script/constants.js
JoshuaKGoldberg, line 1: [Question] Now that the migration tests are all in their own file, is there a reason to split out a constants file?

script/migrate-test-e2e.test.js
JoshuaKGoldberg, line 1: Re the naming of this file: ending with `.test.js` means `vitest` (`pnpm test`) will pick it up. That's why _**Test / test (pull_request)**_ is failing.

I was hopeful [the Vitest CLI's `-t` option](https://vitest.dev/guide/cli) would be useful here... but I tried it out locally and it only seems to match on _test_ names. I don't see any way to specify the [`include` option](https://vitest.dev/config/#include) on the CLI.

Suggestion: how about a `script/vitest.config.ts`? 

```shell
vitest -r script
```

Then that `script/vitest.config.ts` can specify `include: ["migrate-test-e2e.js`]`.

JoshuaKGoldberg: Aside: it'd be nice to have an `--include` option on the Vitest CLI. Maybe a good feature request for them? Thoughts?

JoshuaKGoldberg (changes requested):
I think this is really close! 🙌 Awesome!

script/migrate-test-e2e.test.js
Aslemammad: No they don't have to!

script/migrate-test-e2e.test.js
Aslemammad: I guess having both `--include` and `--exclude` would be ideal! Love that.

package.json
```diff
@@ -36,7 +36,7 @@
 		"test": "vitest",
 		"test:create": "node script/create-test-e2e.js",
 		"test:initialize": "node script/initialize-test-e2e.js",
-		"test:migrate": "node script/migrate-test-e2e.js",
+		"test:migrate": "vitest -r script/",
```
JoshuaKGoldberg, line 39: [Bug] Looks like this is triggering Knip?

https://github.com/JoshuaKGoldberg/create-typescript-app/actions/runs/7313800050/job/19929116425?pr=1048

```shell
$ pnpm lint:knip

> create-typescript-app@1.50.0 lint:knip /home/runner/work/create-typescript-app/create-typescript-app
> knip

The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
Unlisted dependencies (1)
script  package.json
 ELIFECYCLE  Command failed with exit code 1.
Error: Process completed with exit code 1.
```

Might be a bug in Knip. You should be able to ignore it by adding `"ignoreDependencies": ["script"]` in `knip.json`. 

These little things keep popping up haha.

script/__snapshots__/migrate-test-e2e.test.js.snap
JoshuaKGoldberg: bump @Aslemammad, this is still a thing I think?

Aslemammad: oh sorry, i haven't seen this, let me work on it just right now.

Aslemammad:
@JoshuaKGoldberg Let me know if this is ready for a merge.

script/migrate-test-e2e.js
```diff
@@ -1,37 +1,15 @@
...
-);
-if (indexOfUnstagedFilesMessage === -1) {
-	throw new Error(
-		`Looks like migrate didn't cause any file changes? That's ...probably incorrect? 😬`,
-	);
-}
-
 const filesExpectedToBeChanged = new Set([
 	".all-contributorsrc",
-	"bin/index.js",
```
JoshuaKGoldberg, line 30: Hmm, looks like the CI failure is showing this to be changed. So maybe this has to be added back in?

```plaintext
2023-12-30T06:20:05.4901767Z [31m[1m[7m FAIL [27m[22m[39m migrate-test-e2e.js[2m > [22mcheck for unstagedModifiedFiles
2023-12-30T06:20:05.4913830Z [31m[1mError[22m: 
2023-12-30T06:20:05.4914540Z Oh no! Running the migrate script modified some files:
2023-12-30T06:20:05.4915258Z  - bin/index.js
2023-12-30T06:20:05.4915496Z 
2023-12-30T06:20:05.4915948Z That likely indicates changes made to the repository without
2023-12-30T06:20:05.4916809Z corresponding updates to templates in src/.
```

JoshuaKGoldberg (approved):
Wonderful, thanks so much for all your hard work on this PR @Aslemammad! I think it looks great and will be very helpful keeping the last few files aligned to the templates as time goes on. 

I'll apply a couple touchups myself for linking to upstream bugs & adding back `bin/index.js`. But no more blockers. Yay!

Guy with mustache and mullet excitedly saying 'yes that is awesome'

script/__snapshots__/migrate-test-e2e.js.snap
```diff
@@ -0,0 +1,847 @@
...
+-		"test:migrate": "vitest run -r script/",
+ 		"tsc": "tsc"
+ 	},
+ 	"lint-staged": {
+@@ -107,7 +101,7 @@
+ 		"vitest": "^1.0.2",
+ 		"yaml-eslint-parser": "^1.2.2"
+ 	},
+-	"packageManager": "pnpm@8.12.1",
++	"packageManager": "pnpm@8.7.0",
```
JoshuaKGoldberg, line 843: Heh, and in CI:

```plaintext
2023-12-30T06:48:46.9914884Z    	},
2023-12-30T06:48:46.9915527Z - -	"packageManager": "pnpm@8.12.1",
2023-12-30T06:48:46.9916260Z + -	"packageManager": "pnpm@8.13.1",
2023-12-30T06:48:46.9916878Z   +	"packageManager": "pnpm@8.7.0",
```

This'll probably start failing as soon as a new pnpm version is out. Lovely. That can be a followup.

Aslemammad:
> I'll apply a couple touchups myself for linking to upstream bugs & adding back bin/index.js. But no more blockers. Yay!

Thank you so much, happy to see it finally being merged!

script/migrate-test-e2e.js
```diff
@@ -1,37 +1,16 @@
 import chalk from "chalk";
 import { $, execaCommand } from "execa";
+import { test } from "vitest";
+import { assert, expect } from "vitest";
```
JoshuaKGoldberg, line 4: Oop, just noticed:

```suggestion
import { assert, expect, test } from "vitest";
```

JoshuaKGoldberg:
> ❌ codecov/project — 94.27% (-0.02%) compared to 2f8e90f

...very amusing.

JoshuaKGoldberg:
Oh, and: renaming the PR title to `chore:` because this is just an internal-only improvement. It doesn't impact end-users.

JoshuaKGoldberg:
@all-contributors please add @aslemammad for code.

> 🤖 Beep boop! This comment was added automatically by [all-contributors-auto-action](https://github.com/marketplace/actions/all-contributors-auto-action).
> Not all contributions can be detected from Git & GitHub alone. Please comment any missing contribution types this bot missed.
> ...and of course, thank you for contributing! 💙

script/migrate-test-e2e.test.js
JoshuaKGoldberg: Filed: https://github.com/vitest-dev/vitest/issues/4861

PR #1073, opened 2023-12-03
fix: Owner being replaced in github actions

### LINKED ISSUE #1043: 🐛 Bug: Migrating a non-JoshuaKGoldberg repo creates the wrong owners for JoshuaKGoldberg actions
### Bug Report Checklist

### Expected

Running `--mode migrate` on an existing repo under a owner other than `JoshuaKGoldberg` should still set up actions like `JoshuaKGoldberg/release-it-action` under the proper owners.

### Actual

I'd missed in https://github.com/omnidan/node-emoji/pull/142 that two actions were set up incorrectly:

* `omnidan/all-contributors-auto-action` was configured instead of `JoshuaKGoldberg/all-contributors-auto-action`
* `omnidan/release-it-action` was configured instead of `JoshuaKGoldberg/release-it-action`

### Additional Info

_No response_

jaas666:
## PR Checklist

- [X] Addresses an existing open issue: fixes #1043 
- [X] That issue was marked as [`status: accepting prs`](https://github.com/JoshuaKGoldberg/create-typescript-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [X] Steps in [CONTRIBUTING.md](https://github.com/JoshuaKGoldberg/create-typescript-app/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

This PR should fix #1043 Originally I wanted to go with a callback function to exclude certain strings from being replaced but that's above my pay grade.

I don't like the way we are excluding using regex but this should solve the issue in the meantime.

Someday I'll get to that level. 🤓


jaas666:
Yikes.. I'll need help understanding why it fails on the initialization script test.

JoshuaKGoldberg (changes requested):
Yeah I see how you ended up here but:

* This isn't super scalable: there are likely going to be more and more actions over time.
* There's an action mentioned in the issue and not covered here, `release-it-action`

I think the callback method -or something else that doesn't create  long regex- would be the way to go.

jaas666:
I agree. I don't like this solution either.

I tried coming up with a solution using the callback which should work but I couldn't figure out the correct regex to use.

I know my problem was that I change the regex to match `/JoshuaKGoldberg/g` only so the function was always false.

This , along with a new `excludedFromReplacements` variable, was my original attempt. A clever regex should make this work right?

```
[
/JoshuaKGoldberg/g,(match: string) =>
excludedFromReplacements.some((e) =>
match.includes(e))
? Match
: options.owner,
]
```

jaas666:
Sorry. I get and F on this one. I've tried but got nowhere. Feel free to close this PR

JoshuaKGoldberg:
You got it - thanks for trying though! I really appreciate you putting the effort in. ❤️

PR #1637, opened 2024-08-15
fix: exclude tests with.npmignore

### LINKED ISSUE #1629: 🚀 Feature: Test files shouldn't be published
### Bug Report Checklist

### Overview

Right now, CTA repos publish all files under `lib/`:

https://github.com/JoshuaKGoldberg/create-typescript-app/blob/9034d25992800bc8f7891f9cac4bfdf1e2f8c6b2/package.json#L17-L23

...including `*.test.*` files. 😬. That's a lot of extra file size on published packages for no real gain. I think `*.test.*` files should be excluded from the published outputs. 

### Additional Info

I'd previously fixed this in 7cdb836250404b8a3048c1ef281ca468320a2e5f, but at some point switched away from `.npmignore` files and lost that.

💖

johnnyreilly:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1629 

## Overview

Doing this on a mobile phone on a train. It can never work, can it?


.npmignore
JoshuaKGoldberg, line 1: This'll need to be added to the files generated in new repos too, I'm afraid. Just adding it here only works for this repo.

JoshuaKGoldberg (approved):
A start! ✨

.npmignore
johnnyreilly: Ah right - makes sense! Whereabouts would that change be made? (I'm heading off on holiday today so I may have to address this when I return)

JoshuaKGoldberg: [`src/steps/writing/creation/rootFiles.ts`](https://github.com/JoshuaKGoldberg/create-typescript-app/blob/d2f6a1f89389d5ea81581c004034329ff84cafb8/src/steps/writing/creation/rootFiles.ts) is a good starter. `.gitignore` is a rough equivalent, so you could search for that & `createDotGitignore`.

JoshuaKGoldberg (changes requested):
(wrong button)

.npmignore
johnnyreilly: Thanks! Will check in when back from hols

johnnyreilly:
Okay think this is good!

JoshuaKGoldberg (approved):
Yup, this is great - thanks!

I was actually a little surprised at first it's passing build, there are no unit test changes... But all is well:
* Unit tests: I never added snapshots for `rootFiles.ts` because they'd be super verbose.
* End-to-end tests: the migration one is verifying that the `.npmignore` on disk matches what's created

✅ great!

PR #1975, opened 2025-03-09
fix(husky): use pnpx

xav-ie:
Defaulting to npx causes users to unexpectedly also have npx installed. It caused my commit staging to break.

## PR Checklist

- [ ] Addresses an existing open issue: fixes #000
- [ ] That issue was marked as [`status: accepting prs`](https://github.com/JoshuaKGoldberg/create-typescript-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)

## Overview

Simply replaces the `npx lint-staged` command to use `pnpx lint-staged`.

Please, let me know if I need to scale back my changes. I am trying to keep this PR minimal and just get the project working. It was not yet working for me with this dev environment:

```nix
{ ... }: {
  languages.javascript = {
    enable = true;
    # I should not have to enable `npm`, we use `pnpm` in this house!
    # npm.enable = true;
    pnpm.enable = true;
  };
}
```

I have a follow-up PR here:
https://github.com/JoshuaKGoldberg/create-typescript-app/pull/1976
It goes "full-pnpx" because I don't think we should expect users to have both `npx` and `pnpx` installed. `pnpx` in my experience is also way faster since it copies from global store, just like Nix :)))))

It also seems similar in spirit to this PR:
https://github.com/JoshuaKGoldberg/create-typescript-app/pull/1800

I am not sure why the "full-pnpx" move was not taken. Maybe I can learn why not, or just oversight.

🐢

### DIFF
--- a/docs/FAQs.md
+++ b/docs/FAQs.md
@@ -93,7 +93,7 @@ Here we'll outline the steps required to migrate a CTA app to a GitHub Action:
    ```diff
    +pnpm run build
    +git add dist
-   npx lint-staged
+   pnpx lint-staged
    ```
 
 2. Create an [`action.yml` metadata file](https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action#creating-an-action-metadata-file).
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-npx lint-staged
+pnpx lint-staged
--- a/src/blocks/blockPrettier.test.ts
+++ b/src/blocks/blockPrettier.test.ts
@@ -91,7 +91,7 @@ describe("blockPrettier", () => {
 			      ".gitignore": "_
 			",
 			      "pre-commit": [
-			        "npx lint-staged
+			        "pnpx lint-staged
 			",
 			        {
 			          "executable": true,
@@ -203,7 +203,7 @@ describe("blockPrettier", () => {
 			      ".gitignore": "_
 			",
 			      "pre-commit": [
-			        "npx lint-staged
+			        "pnpx lint-staged
 			",
 			        {
 			          "executable": true,
@@ -333,7 +333,7 @@ describe("blockPrettier", () => {
 			      ".gitignore": "_
 			",
 			      "pre-commit": [
-			        "npx lint-staged
+			        "pnpx lint-staged
 			",
 			        {
 			          "executable": true,
--- a/src/blocks/blockPrettier.ts
+++ b/src/blocks/blockPrettier.ts
@@ -85,7 +85,7 @@ pnpm format --write
 			files: {
 				".husky": {
 					".gitignore": "_\n",
-					"pre-commit": ["npx lint-staged\n", { executable: true }],
+					"pre-commit": ["pnpx lint-staged\n", { executable: true }],
 				},
 				".prettierignore": formatIgnoreFile(
 					["/.husky", "/lib", "/pnpm-lock.yaml", ...ignores].sort(),


docs/FAQs.md
```diff
@@ -93,7 +93,7 @@ Here we'll outline the steps required to migrate a CTA app to a GitHub Action:
    ```diff
    +pnpm run build
    +git add dist
-   npx lint-staged
```
xav-ie, line 96: This *could* be removed, but I kept this change because `pnpx` is just better. It's faster and what the project will be using after this commit.

docs/FAQs.md
```diff
@@ -93,7 +93,7 @@ Here we'll outline the steps required to migrate a CTA app to a GitHub Action:
...
+   pnpx lint-staged
```
xav-ie, line 96: This *could* be reverted, but I kept this change because `pnpx` is just better. It's faster and what the project will be using after this commit.

JoshuaKGoldberg: > faster

Is it, though? Trying out an arbitrary text addition of `\nExample.\n` to the bottom of `README.md`:

```plaintext
joshgoldberg ~/repos/create-typescript-app $ hyperfine "npx lint-staged" --warmup 3
Benchmark 1: npx lint-staged
  Time (mean ± σ):     843.7 ms ±   9.0 ms    [User: 833.9 ms, System: 206.5 ms]
  Range (min … max):   832.7 ms … 864.0 ms    10 runs
 
joshgoldberg ~/repos/create-typescript-app $ hyperfine "pnpx lint-staged" --warmup 3
Benchmark 1: pnpx lint-staged
  Time (mean ± σ):     974.1 ms ±  22.9 ms    [User: 854.3 ms, System: 185.1 ms]
  Range (min … max):   943.5 ms … 1007.2 ms    10 runs
```

System info:
- OS: macOS Ventura 13.2
- Hardware: Mac Studio, Apple M1 Max, 32 GB
- Node: 22.12.0
- npm: 10.9.0
- pnpm: 10.5.2

JoshuaKGoldberg (changes requested):
Thanks for looking into this @xav-ie! This is the first time I've heard of a system having `pnpx` but not having `npx`. But I also think you're the first (known) power Nix user I've gotten an issue or PR from 😄.

Request: could you please file an issue with a reproduction I can play with? I'm not a Nix user myself and don't even know how to make a system with Node and not npm.

If `pnpx` really is a little slower than `npx`, my hunch is we'll probably want to turn this into an option that defaults to whichever one is installed on the user's computer. I'm under the impression `npx` is generally meant to be assumed as existing for general Node systems? But maybe that's incorrect and we should reduce the number of dependencies to just have `pnpx`?

docs/FAQs.md
xav-ie: It is not! My mistake!
```sh
> hyperfine "npx lint-staged" "pnpx lint-staged" --warmup 3
Benchmark 1: npx lint-staged
  Time (mean ± σ):     691.0 ms ±   6.2 ms    [User: 815.1 ms, System: 161.2 ms]
  Range (min … max):   685.7 ms … 706.3 ms    10 runs

Benchmark 2: pnpx lint-staged
  Time (mean ± σ):      1.098 s ±  0.047 s    [User: 0.879 s, System: 0.152 s]
  Range (min … max):    1.054 s …  1.184 s    10 runs

Summary
  npx lint-staged ran
    1.59 ± 0.07 times faster than pnpx lint-staged
```

JoshuaKGoldberg:
TIL, thanks for looking into this!

Issue #1991, opened 2025-03-18
🛠 Tooling: Figure out how to test email transforms


JoshuaKGoldberg:
### Tooling Report Checklist

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have pulled the latest `main` branch of the repository.
- [x] I have [searched for related issues](https://github.com/JoshuaKGoldberg/create-typescript-app/issues?q=is%3Aissue) and found none that matched my issue.

### Overview

One of the few still-untested pieces of logic in CTA is the transform of email on the CLI:

https://github.com/JoshuaKGoldberg/create-typescript-app/blob/666e3fa2d1d3b46bc32fac0b9bd5b54f24b7e808/src/base.ts#L72-L83

I don't know how to test this with Bingo. But it'd be good to get this tested.

### Additional Info

Blocked on upstream issue to figure this out: _(todo, will file)_

🎁

PR #2445, opened 2026-08-18
fix: apply --email CLI flag to package.json author

### LINKED ISSUE #2435: 🐛 Bug: --email CLI flag isn't applied to package.json author
### Bug Report Checklist

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have pulled in the newest version of the project.
- [x] I have [searched for related issues](https://github.com/JoshuaKGoldberg/create-typescript-app/issues?q=is%3Aissue) and found none that matched my issue.

### Expected

Passing `--email` on the CLI should populate `package.json`'s `author.email`, per the flag's description ("Email address to be listed as the point of contact in docs and packages") and [`blockPackageJson`](https://github.com/JoshuaKGoldberg/create-typescript-app/blob/main/src/blocks/blockPackageJson.ts#L47):

```ts
author: { email: options.email.npm, name: options.author },
```

### Actual

`author.email` is missing entirely.

```shell
npx create-typescript-app \
  --preset minimal \
  --directory email-repro \
  --description "A repro. 🐛" \
  --owner JoshuaKGoldberg \
  --author JoshuaKGoldberg \
  --email test-flag@example.com
```

```json
"author": { "name": "JoshuaKGoldberg" }
```

Reproduced on `create-typescript-app@2.64.0`, Node 24.13.1, macOS. Also reproduced with `--preset everything`.

### Additional Info

The `--email` string → `{ github, npm }` transform in `base.ts` looks like the suspect:

https://github.com/JoshuaKGoldberg/create-typescript-app/blob/main/src/base.ts#L68-L83

That's the same transform #1991 flags as untested, with a `// TODO: Test this? Is it still working?` comment right above it. This issue is the "is it still working?" half answered: no. #1991 can stay focused on how to test it.

🎁

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2435

## Overview

`--email` was silently dropped on the CLI because of the `z.transform()` on the `email` option:

1. Bingo can't map a `ZodEffects` to a `parseArgs` flag type, so the value became a stray positional.
2. Bingo never `.parse()`s options through their schemas, so the transform never ran anyway.

Replaces the transform with a plain union and a `resolveEmails` helper at the consuming blocks.

The `// TODO` pointing at #1991 is still there. That issue is still open and blocked upstream.

🎁

### DIFF
--- a/src/base.test.ts
+++ b/src/base.test.ts
@@ -6,6 +6,14 @@ import { base } from "./base.js";
 import { AllContributorsData } from "./types.js";
 
 describe("base", () => {
+	// Bingo's CLI can only map an option to a --flag it knows the type of.
+	// A z.transform() hides the string, so --email would be parsed as a boolean.
+	test("email option schema accepts a plain string", () => {
+		const result = base.options.email.safeParse("test@example.com");
+
+		expect(result).toEqual({ data: "test@example.com", success: true });
+	});
+
 	test("production from create-typescript-app", async () => {
 		const options = await prepareOptions(base);
 
--- a/src/utils/resolveEmails.test.ts
+++ b/src/utils/resolveEmails.test.ts
@@ -1,0 +1,26 @@
+import { describe, expect, test } from "vitest";
+
+import { resolveEmails } from "./resolveEmails.js";
+
+describe(resolveEmails, () => {
+	test("string", () => {
+		const actual = resolveEmails("test@example.com");
+
+		expect(actual).toEqual({
+			github: "test@example.com",
+			npm: "test@example.com",
+		});
+	});
+
+	test("object", () => {
+		const actual = resolveEmails({
+			github: "github@example.com",
+			npm: "npm@example.com",
+		});
+
+		expect(actual).toEqual({
+			github: "github@example.com",
+			npm: "npm@example.com",
+		});
+	});
+});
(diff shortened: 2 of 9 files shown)


src/utils/resolveEmails.test.ts
```diff
@@ -0,0 +1,26 @@
+import { describe, expect, test } from "vitest";
+
+import { resolveEmails } from "./resolveEmails.js";
+
+describe(resolveEmails, () => {
+	test("string", () => {
```
JoshuaKGoldberg: These test titles should look like it("X when Y")

src/base.test.ts
```diff
@@ -6,6 +6,14 @@ import { base } from "./base.js";
...
 
 describe("base", () => {
+	// Bingo's CLI can only map an option to a --flag it knows the type of.
+	// A z.transform() hides the string, so --email would be parsed as a boolean.
+	test("email option schema accepts a plain string", () => {
+		const result = base.options.email.safeParse("test@example.com");
+
+		expect(result).toEqual({ data: "test@example.com", success: true });
+	});
+
```
JoshuaKGoldberg:
```suggestion
```

This test doesn't belong here, it's email-specific. If it doesn't have a clear better place, just remove it.

PR #2469, opened 2026-09-19
feat: generate the bin file when --bin is set

### LINKED ISSUE #2468: 🚀 Feature: Generate the bin file when --bin is set
- [x] I have looked at the latest version of the project.
- [x] I have [searched for related issues](https://github.com/JoshuaKGoldberg/create-typescript-app/issues?q=is%3Aissue) and found none that matched my issue.

## Overview

When `--bin` is set, generate the bin file (e.g. `bin/index.js`) as a CTA-owned file instead of asking users to create it by hand.
Intake should preserve an existing bin file's contents, so transition mode doesn't clobber custom logic.

## Additional Info

Prerequisite for migrating existing repositories in #2462: once CTA owns the file, its `../lib/` import can be rewritten during intake instead of by a shell script.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2468

## Overview

Adds `blockBin`, a preserve-and-migrate block for the primary `--bin` file:

- **Existing repositories**: intake reads the current bin file and re-emits it verbatim. An audit of the 16 CTA-templated repos under @JoshuaKGoldberg with a `bin` found no two alike (different entry paths, export names, argv and error handling, one self-contained CLI, one bin inside build output), so CTA never generates over a user's file. The one side effect is `executable: true`, so a transition may add a mode change where the file isn't already `+x`.
- **New repositories** (`--mode create`): a starter file is generated at the bin path importing the built entry (`../lib/index.mjs`, depth-adjusted), replacing the FAQ's "create the file yourself". A `contents` addon lets a repo supply its own; this repo does in the integration test.

The block is in every preset so `--bin` works regardless of preset, but it emits nothing unless `--bin` is set or `package.json` already has a `bin`. `docs/Blocks.md` and the bin FAQ are updated.

Prerequisite for #2467: owning the file's emission is what lets the `lib/` → `dist/` migration rewrite `../lib/` imports during intake rather than with a shell script.

🎁


src/blocks/blockBin.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const blockBin = base.createBlock({
+	about: {
+		name: "Bin",
+	},
+	addons: {
+		contents: z.string().optional(),
+	},
+	// Bin files are bespoke per repository, so an existing one is preserved as-is.
+	// The starter contents below only apply when creating a new repository.
```
JoshuaKGoldberg:
```suggestion
```

src/integration.test.ts
```diff
@@ -48,6 +49,15 @@ test("Producing the everything preset matches the files in this repository", asy
...
 			addons: [
+				blockBin({
+					contents: `#!/usr/bin/env node
+import { runTemplateCLI } from "bingo";
+
+import { template } from "../lib/index.mjs";
+
+process.exitCode = await runTemplateCLI(template);
+`,
+				}),
```
JoshuaKGoldberg: shouldn't this infer?

JoshuaKGoldberg: It does now: b82da0f1 passes the intaken files to producePreset, so blockBin (and Codecov, CSpell, Knip, Renovate) infer from disk and the explicit contents are gone.

src/integration.test.ts
```diff
@@ -33,9 +29,9 @@ vi.mock("./utils/resolveBin.ts", () => ({
 //
 // The next most likely culprit for failures is changing file contents that are
 // specified by the addons mentioned in the producePreset() call below.
-// For now, if you change the output on disk, you'll need to manually update here too.
-// TODO: Eventually the create engine will be able to infer them:
-//   https://github.com/JoshuaKGoldberg/bingo/issues/128
+// Most addons are inferred from the files on disk by each block's intake();
+// only the ones without an intake need to be listed here.
+// If you change one of those on disk, you'll need to manually update here too.
 //
```
JoshuaKGoldberg, line 37:
```suggestion
// TODO: Eventually the create engine will be able to infer them:
//   https://github.com/JoshuaKGoldberg/bingo/issues/128
//
```

